PR #21, opened 2021-06-10
[BD-26] Improve handling of API errors, refactor redux store initialization, fix multiple API calls, add new pages

ViktorRusakov:
1) Cover all API calls with error handler, updated error component with show / hide buttons, add decisions file
2) Refactor redux store initialization so that it contains some example data instead of empty objects
3) Refactor API calls related to proctoring because they used to get called multiple times
4) Added expired pages for timed / onboarding / proctored exams and onboarding failed pages for proctored exams.


edx-webhook:
Thanks for the pull request, @ViktorRusakov! I've created [BLENDED-882](https://openedx.atlassian.net/browse/BLENDED-882) to keep track of it in Jira.   More details are on the [BD-26](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2523627980/BD-26+Proctoring+MFE+and+configuration) project page.


When this pull request is ready, tag your edX technical lead.

src/constants.js
```diff
@@ -41,3 +58,5 @@ export const ExamType = Object.freeze({
   PROCTORED: 'proctored',
   TIMED: 'timed',
 });
+
+export const NON_PRACTICE_EXAMS = [ExamType.ONBOARDING, ExamType.PROCTORED, ExamType.TIMED];
```
Zacharis278, line 62: this is somewhat confusing because an onboarding exam is technically a practice exam. This appears to only get used in a single test anyway, maybe let's just explicitly list these states there

src/instructions/index.jsx
```diff
@@ -1,37 +1,55 @@
...
+  const { exam, verification } = state;
   const { attempt, type: examType, prerequisite_status: prerequisitesData } = exam || {};
   const prerequisitesPassed = prerequisitesData ? prerequisitesData.are_prerequisites_satisifed : true;
   let verificationStatus = verification.status || '';
   const { verification_url: verificationUrl } = attempt || {};
   const [skipProctoring, toggleSkipProctoring] = useState(false);
   const toggleSkipProctoredExam = () => toggleSkipProctoring(!skipProctoring);
+  const expired = shouldRenderExpiredPage(exam);
+
+  if (expired) {
```
Zacharis278, line 38: this file is way easier to read, thanks!

src/exam/ExamAPIError.jsx
```diff
@@ -1,38 +1,66 @@
...
+              + 'Please reach out to {supportLink} for assistance, '
               + 'and return to the exam once you receive further instructions.'
           }
-          values={{ support_link: <a href={contactUs} target="_blank" rel="noopener noreferrer">{platformName} Support</a> }}
+          values={{ supportLink: <a href={contactUs} target="_blank" rel="noopener noreferrer">{platformName} Support</a> }}
         />
+      );
+    }
+
+    return (
```
Zacharis278, line 28: do we have tests for this error component?

ViktorRusakov: no, we don't
added them in 906b579

src/constants.js
ViktorRusakov: yes, I see your point, didn't think of it that way, removed it in 906b579

Zacharis278 (approved):

PR #38, opened 2021-07-26
[BD-26] Add handling of second review required status for proctored exams

ViktorRusakov:
Fix missing second_review_required status for proctored exams, which resulted in showing exam content to the student

JIRA: [EDUCATOR-5909](https://openedx.atlassian.net/browse/EDUCATOR-5909)

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -7,6 +7,7 @@ export const ExamStatus = Object.freeze({
   STARTED: 'started',
   READY_TO_SUBMIT: 'ready_to_submit',
   SUBMITTED: 'submitted',
+  SECOND_REVIEW_REQUIRED: 'second_review_required',
   TIMED_OUT: 'timed_out',
   VERIFIED: 'verified',
   REJECTED: 'rejected',
--- a/src/exam/ExamAPIError.jsx
+++ b/src/exam/ExamAPIError.jsx
@@ -27,7 +27,7 @@ export default function ExamAPIError() {
 
     return (
       <FormattedMessage
-        id="exam.apiError.text2"
+        id="exam.defaultError"
         defaultMessage="A system error has occurred with your exam. Please reach out to support for assistance."
       />
     );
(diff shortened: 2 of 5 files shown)


edx-webhook:
Thanks for the pull request, @ViktorRusakov! I've created [BLENDED-928](https://openedx.atlassian.net/browse/BLENDED-928) to keep track of it in Jira.   More details are on the [BD-26](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2523627980/BD-26+Proctoring+MFE+and+configuration) project page.


When this pull request is ready, tag your edX technical lead.

ViktorRusakov:
@Zacharis278 I noticed that in LMS this attempt status (and some others) are handled [differently](https://github.com/edx/edx-proctoring/blob/68827c066e3afcbc3fc57ae3c991077f876b53b2/edx_proctoring/api.py#L2745) for proctored exams

do we also need to add that check which uses `_was_review_status_acknowledged` function to our API? I don't think we do that check right now

idegtiarov (approved):
LGTM!

UvgenGen (approved):

Zacharis278:
this bug has me thinking that we should be more strict about rendering the `children` in this function. If someone added a new backend attempt state it would automatically show the exam content. Can we make that case more specific to only render content for a started attempt or a timed exam that is viewable after submission?

ViktorRusakov:
Yeah, we can do that, but what should the default behaviour be then? What do we show if unknown status came from the backend?

Zacharis278:
@ViktorRusakov I think we can just add a simple error template. We don't really expect to ever hit this case unless something has gone wrong

ViktorRusakov:
@Zacharis278 I've updated the PR to handle unknown statuses, please take a look

idegtiarov (approved):
lgtm

Zacharis278 (approved):
thanks! This looks good, can you queue up the version bump for the learning MFE as well?

ViktorRusakov:
sure, I've opened new PR [here](https://github.com/edx/frontend-app-learning/pull/564)

waiting for this to be merged and I'll update it then

PR #39, opened 2021-08-03
feat: allow staff users to bypass exam interstitials

Zacharis278:
## [MST-943](https://openedx.atlassian.net/browse/MST-943)

Staff users should be able to interact with exam content without going through the exam flow


schenedx (approved):
👍

PR #40, opened 2021-08-23
fix: display exam content for masquerading user

bseverino:
[MST-944](https://openedx.atlassian.net/browse/MST-944)

If a staff user is masquerading as a specific learner, they should be able to view exam content regardless of the learner's state. If the learner would not normally be able to see the exam content, an alert with this information will display for the masquerading user.

Related PRs:
https://github.com/edx/edx-platform/pull/28458
https://github.com/edx/frontend-app-learning/pull/610

---

### Staff view:
Screen Shot 2021-08-23 at 1 18 06 PM

---

### Learner view:
Screen Shot 2021-08-23 at 1 24 26 PM

### DIFF
--- a/src/exam/Exam.jsx
+++ b/src/exam/Exam.jsx
@@ -1,11 +1,13 @@
 import React, { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { Spinner } from '@edx/paragon';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { Alert, Spinner } from '@edx/paragon';
+import { Info } from '@edx/paragon/icons';
 import { ExamTimerBlock } from '../timer';
 import Instructions from '../instructions';
 import ExamStateContext from '../context';
 import ExamAPIError from './ExamAPIError';
-import { ExamType } from '../constants';
+import { ExamStatus, ExamType } from '../constants';
 
 /**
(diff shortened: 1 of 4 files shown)


Zacharis278:
for staff users who are masquerading, is there no way to tell they're staff just by looking at `authenticatedUser`?

src/exam/Exam.jsx
```diff
@@ -51,6 +74,14 @@ const Exam = ({ isTimeLimited, children }) => {
 
   return (
     <div className="d-flex flex-column justify-content-center">
+      {shouldShowMasqueradeAlert() && (
+        <Alert variant="info" icon={Info} data-testid="masquerade-alert">
+          <FormattedMessage
+            id="exam.hiddenContent"
```
schenedx, line 80: Do we need this message to be specified somewhere for translation?

bseverino:
> for staff users who are masquerading, is there no way to tell they're staff just by looking at `authenticatedUser`?

`authenticatedUser` has an `administrator` value, but I don't think we can account for course staff in that case

src/exam/Exam.jsx
bseverino: Nope, `FormattedMessage` takes care of that

schenedx (approved):
👍

Zacharis278 (approved):

PR #41, opened 2021-08-30
feat: support for content gating

Zacharis278:
### [MST-991](https://openedx.atlassian.net/browse/MST-991)

Handle content gating by allowing default learning MFE rendering instead of exam interstitials for sections that are locked. Note this behavior is distinctly different from the exam prerequisite views which are based solely on CreditRequirements. In the case both are not satisfied the effect of this PR is that content gating would take precedence. This is how the old LMS views currently work.

Bonus: Fixed some HTML validation that was barfing errors into the test console

@edx/masters-devs-cosmonauts


schenedx (approved):
👍

PR #43, opened 2021-09-22
[fix] The special exam timer synced with backend to show accurate count down timer

schenedx:
[MST-978 ](https://openedx.atlassian.net/browse/MST-978)

The special exam timer shown to learners should be updated with the exam attempt remaining time available for each sync interval. This is to ensure our learners don't see a timer which is running slower than the actual time.

Please review @edx/masters-devs-cosmonauts

### DIFF
--- a/src/timer/TimerProvider.jsx
+++ b/src/timer/TimerProvider.jsx
@@ -39,8 +39,8 @@ const TimerServiceProvider = ({
     critically_low_threshold_sec: criticalLowTime,
     low_threshold_sec: lowTime,
   } = attempt;
-  const startValue = Math.floor(timeRemaining);
   const LIMIT = GRACE_PERIOD_SECS ? 0 - GRACE_PERIOD_SECS : 0;
+  let liveInterval = null;
 
   const getTimeString = () => Object.values(timeState).map(
     item => {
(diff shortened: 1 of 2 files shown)


src/timer/TimerProvider.jsx
```diff
@@ -77,26 +77,34 @@ const TimerServiceProvider = ({
...
-        pollExam();
-      }
-
-      // if exam is proctored ping provider app also
-      if (workerUrl && timerTick % pingInterval === pingInterval / 2) {
-        pingHandler(pingInterval, workerUrl);
+    if (liveInterval) {
+      clearInterval(liveInterval);
+      liveInterval = null;
+    } else {
```
Zacharis278, line 84: can you explain why this is in an else statement? Don't we want to create a new timer after clearing the old one?

schenedx: You are right. That's a mistake I made here. Thank you for pointing that out. Will update

schenedx: Updated. @Zacharis278 Please review again.

Zacharis278 (approved):
lgtm. It seems to me your test would actually cover L82-83 codecov is yelling about 🤔

PR #44, opened 2021-09-23
Revert "[fix] The special exam timer synced with backend to show accurate count down timer (#43)

schenedx:
This reverts commit b34e0f921e8819a1e7ff1b45687ab0d8c3a8cc24.

The commit message is not correct, and it didn't trigger a version update. Revert and then re-introduce with the correct commit message.


Zacharis278 (approved):
seems like this creates a messier history than just pushing a followup fix commit doesn't it? Up to you tho

PR #46, opened 2021-09-23
build: integrate commitlint

schenedx:
Integrate commitlint into the repository pre-commit check so commit messages can trigger semantic-release properly.

See example failed commit messaging below:
Screen Shot 2021-09-23 at 3 32 17 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,8 @@ jobs:
       run: npm run lint
     - name: Test
       run: npm run test
+    - name: commit-lint
+      run: git log -1 --pretty=format:"%s" | npx commitlint
     - name: Build
       run: npm run build
     - name: i18n_extract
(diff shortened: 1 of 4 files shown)


ashultz0 (approved):

Zacharis278:
a thought, would the be better to attach to the github CI instead of running on commit which can be ignored?

schenedx:
@Zacharis278 Running commitlint on github CI sounds like a good idea. Why not have both?

Zacharis278:
both sounds fine. What I am realizing is that this isn't foolproof since ultimately the commit that goes into master comes from the github webui. 🙃

schenedx:
@Zacharis278 I guess. But the enforcement are here to make it difficult to make mistake. That should be enough.

Zacharis278 (approved):

PR #52, opened 2021-11-05
[MST-1150] Don't show IDV interstitial if IDV is disabled for the course

bseverino:
[MST-1150](https://openedx.atlassian.net/browse/MST-1150)

IDV is considered disabled for a course if the integrity signature flag is enabled. If this flag is turned on, do not block the learner from taking proctored exams if they do not have approved ID verification.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/frontend-lib-special-exams",
-  "version": "1.13.0",
+  "version": "1.14.0",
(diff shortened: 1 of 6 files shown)


alangsto (approved):
These look good to me! Is there a place to add a unit test/would it be worth it? I know we're going to take this code out later.

bseverino:
@alangsto Yeah good call, it was easy enough to add.

PR #53, opened 2021-12-06
fix: log worker errors for startProctoredExam

bseverino:
[MST-1162](https://openedx.atlassian.net/browse/MST-1162)

### DIFF
--- a/src/data/thunks.js
+++ b/src/data/thunks.js
@@ -1,4 +1,4 @@
-import { logError } from '@edx/frontend-platform/logging';
+import { logError, logInfo } from '@edx/frontend-platform/logging';
 import {
   fetchExamAttemptsData,
   createExamAttempt,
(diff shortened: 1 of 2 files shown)


src/data/thunks.js
```diff
@@ -150,10 +150,22 @@ export function startProctoredExam() {
         .then(() => updateAttemptAfter(
           exam.course_id, exam.content_id, continueAttempt(attempt.attempt_id),
         )(dispatch))
-        .catch(() => handleAPIError(
-          { message: 'Something has gone wrong starting your exam. Please double-check that the application is running.' },
-          dispatch,
-        ));
+        .catch(error => {
+          if (error) {
+            logInfo(
```
schenedx, line 155: Why logInfo instead of just logError here?

schenedx (commented):
Got a question. Nice test setup and checking!

src/data/thunks.js
bseverino: I thought about it initially, but I'm afraid it'll blow up the error rate, especially if there is user error involved.

Zacharis278 (approved):
not totally familiar with how our frontend logging works, can you brief me how we can find these?

bseverino:
@Zacharis278 Yeah, there's a pretty good overview here: https://github.com/edx/frontend-platform/blob/4c9057b5796a8cf9090eb0f020904c2ef547c1cc/src/logging/NewRelicLoggingService.js#L43

src/data/thunks.js
schenedx: It's a fair concern. I still think we should do logError. I thought we can just exclude this one on newrelic? Or at least set the alert threshold high?
To me, this is a good error to alert us anyways to actively help and identify learner issues

schenedx (approved):
Either way, I am good with this change.

PR #54, opened 2021-12-22
Improve Exam Start Failure

Zacharis278:
1. The application will now wait 5 seconds to connect with the Proctortrack application before showing an error. This previously took two minutes with no user feedback what was happening.  Under the hood this will retry five times still.
2. Added a spinner to the button while this is taking place
3. Changed to the log level for these errors from info to error. Info messages don't seem to be getting logged, not sure why. Either way this should have no significant impact on error rates as we receive up to 600/hr for translation errors alone.

Note: This can be merged, but the timeout will still be two minutes until https://github.com/anupdhabarde/edx-proctoring-proctortrack/pull/19 is also merged

@edx/masters-devs-cosmonauts 

[MST-1252](https://openedx.atlassian.net/browse/MST-1252)

Screen Shot 2021-12-22 at 11 26 13 AM


schenedx (approved):
👍

PR #56, opened 2022-01-13
fix: exam id logging

Zacharis278:
The attempt object does not actually have an exam id on it so this was blowing up. Didn't get caught because we didn't have any testing on ping weirdly, now we do.

Also homogenized the log format between ping failures and start exam failures and included exam states so we can better differentiate in newrelic


schenedx (approved):
👍

PR #58, opened 2022-02-08
[MST-1266] Improve API error messaging

bseverino:
https://openedx.atlassian.net/browse/MST-1266

Display the relevant error message more prominently, and fall back to a default message if there is none, or if the message is HTML. Also use the values defined in frontend config for site name + support URL, rather than relying on proctored exam settings.

Before
-------
43aaaade-4a3d-48ee-bebe-84ee48b954fe

After
-----

**Using the API error message:**
Screen Shot 2022-02-08 at 4 40 12 PM

**Default message:**
Screen Shot 2022-02-08 at 4 39 21 PM

### DIFF
--- a/src/exam/ExamAPIError.jsx
+++ b/src/exam/ExamAPIError.jsx
@@ -1,66 +1,51 @@
-import React, { useState, useContext } from 'react';
-import { Alert, Icon } from '@edx/paragon';
+import React, { useContext } from 'react';
+import { getConfig } from '@edx/frontend-platform';
+import { Alert, Hyperlink, Icon } from '@edx/paragon';
 import { Info } from '@edx/paragon/icons';
-import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { injectIntl, intlShape, FormattedMessage } from '@edx/frontend-platform/i18n';
 import ExamStateContext from '../context';
+import messages from './messages';
 
-export default function ExamAPIError() {
+function ExamAPIError({ intl }) {
   const state = useContext(ExamStateContext);
-  const { apiErrorMsg, proctoringSettings } = state;
-  const { contact_us: contactUs, platform_name: platformName } = proctoringSettings;
-  const [showError, setShowError] = useState(false);
-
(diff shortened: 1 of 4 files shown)


src/exam/ExamAPIError.jsx
```diff
@@ -1,66 +1,59 @@
...
+  const shouldShowApiErrorMsg = !!apiErrorMsg && !apiErrorMsg.includes('<');
+
+  const renderSupportMsg = () => {
+    const { SITE_NAME, SUPPORT_URL } = getConfig();
+
+    if (SITE_NAME && SUPPORT_URL) {
       return (
         <FormattedMessage
-          id="exam.apiError.text1"
+          id="exam.apiError.supportText.one"
```
schenedx, line 21: I am wondering, why we use number on the IDs here? Why not use descriptive text?

src/exam/messages.js
```diff
@@ -0,0 +1,14 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  apiErrorDefault: {
+    id: 'exam.apiError.default',
+    defaultMessage: 'A system error has occurred with your exam.',
+  },
+  supportTextTwo: {
```
schenedx, line 8: Can we avoid using number postfix here? Maybe something more descriptive?

src/exam/ExamAPIError.jsx
```diff
@@ -1,66 +1,59 @@
...
-    );
+    return intl.formatMessage(messages.supportTextTwo);
   };
 
   return (
     <Alert variant="danger" data-testid="exam-api-error-component">
       <Icon src={Info} className="alert-icon" />
-      <Alert.Heading data-testid="heading">
-        {renderHeading()}
+      <Alert.Heading data-testid="error-details">
```
schenedx, line 47: Why not just include the heading rendering also into the function `renderSupportMsg`? It's going to be easier to read, no?

schenedx (approved):
👍

PR #59, opened 2022-02-15
fix: use a new CourseMeta property restrict exam access

schenedx:
[MST-1273](https://openedx.atlassian.net/browse/MST-1273)

We should not display exam content for learners who are in audit or honor enrollment tracks, if the exam type is not timed exams.

@edx/masters-devs-cosmonauts Please review


src/exam/messages.js
```diff
@@ -9,6 +9,10 @@ const messages = defineMessages({
     id: 'exam.apiError.supportText.withoutLink',
     defaultMessage: 'If the issue persists, please reach out to support for assistance, and return to the exam once you receive further instructions.',
   },
+  proctoredExamAccessDenied: {
+    id: 'exam.proctoredExamDenied',
+    defaultMessage: 'Cannot access proctored exams if user not eligible to receive certificates.',
```
bseverino, line 14: I think this can be more user-friendly - `You do not have access to proctored exams with your current enrollment`, or something like that?

bseverino (commented):
The logic looks good to me, just a nit with wording

src/exam/messages.js
schenedx: Sounds good. The message updated according to your suggestion.

bseverino (approved):

src/exam/messages.js
schenedx: The new words screenshot
dontHaveProctoredAccess

PR #60, opened 2022-02-17
fix: only deny audit and honor code learners of certain exam types

schenedx:
### Description
This change fixes a bug associated with version 1.15.4 by [this PR](https://github.com/edx/frontend-lib-special-exams/pull/59).

The bug was that, the [code at frontend-lib-special-exam](https://github.com/edx/frontend-lib-special-exams/blob/main/src/exam/Exam.jsx#L61), which is part of the version 1.15.4, is excluding too much. When the content type is not of an exam type, the variable examType has the value "". I need to only check against an array of proctored exam types.

@edx/masters-devs-cosmonauts Please review

### DIFF
--- a/src/exam/Exam.jsx
+++ b/src/exam/Exam.jsx
@@ -54,11 +54,13 @@ const Exam = ({
 
   const [hasProctoredExamAccess, setHasProctoredExamAccess] = useState(true);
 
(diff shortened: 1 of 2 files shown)


alangsto (approved):
👍

PR #65, opened 2022-04-07
fix: ease peer dependency requirements to just major version

mikix:
The versions were so tightly defined, it made it hard for a consuming MFE to upgrade minor versions of other packages.


alangsto (approved):
👍

PR #66, opened 2022-04-08
fix: update paragon to latest 19.x version

mikix:


package.json
```diff
@@ -66,7 +66,7 @@
   "devDependencies": {
     "@edx/frontend-build": "^5.6.11",
     "@edx/frontend-platform": "1.8.4",
-    "@edx/paragon": "13.17.0",
+    "@edx/paragon": "19.14.1",
```
mikix, line 69: I could not find any relevant breaking changes in 13->19. Plus, the learning MFE has been on 19 for a while while using this lib and we haven't heard reports of issues yet.

src/timer/CountDownTimer.test.jsx
```diff
@@ -58,8 +58,8 @@ describe('ExamTimerBlock', () => {
 
     expect(screen.getByRole('alert')).toBeInTheDocument();
     expect(screen.getByText(attempt.exam_display_name)).toBeInTheDocument();
-    expect(screen.getAllByRole('button').length).toEqual(2);
-    expect(screen.getByRole('button', { name: 'Show more' })).toBeInTheDocument();
+    expect(screen.getByText('Show more')).toBeInTheDocument();
+    expect(screen.getAllByRole('button').length).toEqual(1);
```
mikix, line 62: These 'show more' / 'show less' `Alert.Link` components went from buttons to a link (which you would think I could use `getByRole('link')` to grab, but I think because they don't have an href, they aren't really a11y links).

Zacharis278 (approved):

PR #71, opened 2022-10-11
refactor: migrate off modal paragon deprecated component

Mashal-m:
### Ticket
[Migrate off deprecated Paragon components](https://github.com/openedx/frontend-wg/issues/102)

### What has changed?

- Migrate off paragon modal deprecation in **WarningModal.jsx**


arbrandes:
@leangseu-edx, any chance you could review this?

@Mashal-m,  mind updating the branch with master in the meantime?

leangseu-edx (changes requested):
Shouldn't `ModalDialog.body` is uppercase? 
Anyway, can you use `ModalDialogBody`, ModalDialogHeader`.... syntax instead? Since those are import directly and it is less likely to create typo and it is easier to check with the doc on paragon https://paragon-openedx.netlify.app/components/modal/modal-dialog/

Mashal-m:
> Shouldn't `ModalDialog.body` is uppercase? Anyway, can you use `ModalDialogBody`, ModalDialogHeader`.... syntax instead? Since those are import directly and it is less likely to create typo and it is easier to check with the doc on paragon https://paragon-openedx.netlify.app/components/modal/modal-dialog/

According to paragon documentation, we can only import _ModaDialog_ and _ModalCloseButton_ directly. It is mandatory to use header, title, etc like this _ModalDialog.Body_.

leangseu-edx (approved):
LGTM 👍

leangseu-edx:
> According to paragon documentation, we can only import _ModaDialog_ and _ModalCloseButton_ directly. It is mandatory to use header, title, etc like this _ModalDialog.Body_.

Fair. It seems all of the example are doing that.

PR #72, opened 2022-10-19
feat: pass external attempt id to provider js

Zacharis278:
### [MST-1659](https://2u-internal.atlassian.net/browse/MST-1659)

Supporting PR to: https://github.com/openedx/edx-proctoring/pull/1084

Take the external id from the exam attempt and pass it off to the JS worker


Zacharis278:
we don't have any tests around the js worker so I've just had to manually test this.

alangsto (approved):

PR #82, opened 2023-02-07
feat: fetch exam access token

varshamenon4:
From the Jira ticket, [MST-1599](https://2u-internal.atlassian.net/browse/MST-1599): 

> Before the frontend-lib-special-exams can render any content within an exam, it should first fetch an access ticket. The access ticket will eventually be passed down to the xblock via cookie, but this ticket should only involve making a call to get the access ticket.
> 
> Prior to rendering any content (aka blocking rendering), the exams library should make a request to the get exam access ticket endpoint, and the endpoint will create a ticket and set it as a cookie.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,0 +1,16 @@
+import { examRequiresAccessToken, store } from './data';
+
+export function isExam() {
+  const { exam } = store.getState().examState;
+  return exam.id !== null;
+}
+
+export function getExamAccess() {
+  const { examAccessToken } = store.getState().examState;
+  return examAccessToken.exam_access_token;
+}
+
+export async function fetchExamAccess() {
+  const { dispatch } = store;
+  return dispatch(examRequiresAccessToken());
+}
--- a/src/data/thunks.js
+++ b/src/data/thunks.js
@@ -1,4 +1,5 @@
 import { logError } from '@edx/frontend-platform/logging';
+import { getConfig } from '@edx/frontend-platform';
 import {
   fetchExamAttemptsData,
   fetchLatestAttempt,
@@ -13,6 +14,7 @@ import {
   resetAttempt,
   declineAttempt,
   endExamWithFailure,
+  fetchExamAccessToken,
 } from './api';
 import { isEmpty } from '../helpers';
 import {
@@ -21,6 +23,7 @@ import {
   expireExamAttempt,
   setActiveAttempt,
   setProctoringSettings,
+  setExamAccessToken,
   setReviewPolicy,
   setApiError,
   setAllowProctoringOptOut,
@@ -60,7 +63,6 @@ function updateAttemptAfter(courseId, sequenceId, promiseToBeResolvedFirst = nul
         if (!noLoading) { dispatch(setIsLoading({ isLoading: false })); }
       }
     }
-
     try {
       const attemptData = await fetchExamAttemptsData(courseId, sequenceId);
       dispatch(setExamState({
@@ -112,6 +114,25 @@ export function getProctoringSettings() {
       dispatch(setProctoringSettings({ proctoringSettings }));
     } catch (error) {
       handleAPIError(error, dispatch);
+    }
+  };
+}
+
+export function examRequiresAccessToken() {
+  return async (dispatch, getState) => {
+    if (!getConfig().EXAMS_BASE_URL) {
+      return;
+    }
+    const { exam } = getState().examState;
+    if (!exam.id) {
+      logError('Failed to get exam access token. No exam id.');
+      return;
+    }
+    try {
+      const examAccessToken = await fetchExamAccessToken(exam.id);
+      dispatch(setExamAccessToken({ examAccessToken }));
+    } catch (error) {
+      logError('Exam access token was not granted.');
     }
   };
 }
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -1,2 +1,8 @@
 export { default } from './core/SequenceExamWrapper';
 export { default as OuterExamTimer } from './core/OuterExamTimer';
+export {
+  getExamAccess,
+  isExam,
+  fetchExamAccess,
+} from './api';
+export { store } from './data';
--- a/src/data/__factories__/examAccessToken.factory.js
+++ b/src/data/__factories__/examAccessToken.factory.js
@@ -1,0 +1,7 @@
+import { Factory } from 'rosie'; // eslint-disable-line import/no-extraneous-dependencies
+
+Factory.define('examAccessToken')
+  .attrs({
+    exam_access_token: 'Z173480948902JK34432',
+    exam_access_token_expiration: '60',
+  });
--- a/src/data/__factories__/examState.factory.js
+++ b/src/data/__factories__/examState.factory.js
@@ -2,10 +2,12 @@ import { Factory } from 'rosie'; // eslint-disable-line import/no-extraneous-dep
 
 import './exam.factory';
 import './proctoringSettings.factory';
+import './examAccessToken.factory';
 
 Factory.define('examState')
   .attr('proctoringSettings', Factory.build('proctoringSettings'))
   .attr('exam', Factory.build('exam'))
+  .attr('examAccessToken', Factory.build('examAccessToken'))
   .attrs({
     isLoading: false,
     activeAttempt: null,
--- a/src/data/__factories__/index.js
+++ b/src/data/__factories__/index.js
@@ -2,3 +2,4 @@ import './examState.factory';
 import './exam.factory';
 import './attempt.factory';
 import './proctoringSettings.factory';
+import './examAccessToken.factory';
--- a/src/data/__snapshots__/redux.test.jsx.snap
+++ b/src/data/__snapshots__/redux.test.jsx.snap
@@ -1,5 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Data layer integration tests Test examRequiresAccessToken for exams IDA url Should get exam access token 1`] = `
+Object {
+  "exam_access_token": "Z173480948902JK34432",
+  "exam_access_token_expiration": "60",
+}
+`;
+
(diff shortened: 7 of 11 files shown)


src/data/thunks.js
```diff
@@ -98,6 +97,32 @@ export function getProctoringSettings() {
   };
 }
 
+// todo: should this happen on its own or within updateAttemptAfter or somewhere else?
```
varshamenon4, line 100: Question: what I'm wondering here is if this function should happen within one of the other functions or if it can be standalone as is. My sense is the latter, as it would wait for the fetch access token call. Plus, I'm not sure that updateAttemptAfter would be the best place for fetchExamAccessTicket anyway since we want to also check for exam access even if there is no exam attempt (for past due exams).

src/data/thunks.js
```diff
@@ -98,6 +97,32 @@ export function getProctoringSettings() {
...
+    const { exam } = getState().examState;
+    if (!exam.id) {
+      logError('Failed to get exam settings. No exam id.');
+      handleAPIError(
+        { message: 'Failed to fetch exam access ticket. No exam id was found.' },
+        dispatch,
+      );
+      // return;
+    }
+    // try {
```
varshamenon4, line 113: Note: I commented this out because the commit was failing due to some missing parts, but this is the main chunk where I have questions about the implementation. A review of this currently commented block would be really helpful!

src/data/thunks.js
```diff
@@ -98,6 +97,32 @@ export function getProctoringSettings() {
...
+    //   const examAccessTicketResponse = await fetchExamAccessTicket(exam.id);
+    //   const examAccessTicket = examAccessTicketResponse.cookies.exam_access_ticket;
+    //   // todo: error if access ticket is expired. Is it possible for it to exist if expired?
```
varshamenon4, line 116: My understanding was that the cookie would be deleted if expired, so I wasn't sure if this check was required. But it was something that came up as a potential AC.

src/data/thunks.js
```diff
@@ -98,6 +97,32 @@ export function getProctoringSettings() {
...
+    if (!exam.id) {
+      logError('Failed to get exam settings. No exam id.');
+      handleAPIError(
+        { message: 'Failed to fetch exam access ticket. No exam id was found.' },
+        dispatch,
+      );
+      // return;
+    }
+    // try {
+    //   const examAccessTicketResponse = await fetchExamAccessTicket(exam.id);
```
varshamenon4, line 114: Note: so here, I'm fetching an exam ticket then grabbing the actual cookie.

src/data/thunks.js
MichaelRoytman: How does the cookie get deleted on expiration?

I know that the unpack function in token-utils, which the platform will call, will raise an error if the cookie is expired. Is that what you're referring to? Or is there some other process?

edx-exams will generate a new access token each time the access token API is hit, right? Do we plan on hitting that API only right before the learner enters the exam? In what cases is it probable that the token will have expired? I'd imagine it's very unlikely the token would have expired at this point, right?

src/data/thunks.js
Zacharis278: So with cookies there's no need for the frontend code to manage them. Because the backend code we wrote returns a set-cookie response the browser will do everything for us as far as storing it, handling expiration, and sending along to any request to *.edx.org.

MichaelRoytman (approved):

src/data/thunks.js
MichaelRoytman: Ah, I just saw Zach's comment. This may be moot then.

alangsto: Can the backend return an expired access ticket? If yes, we would want to handle that case.

MichaelRoytman:
Whoops! I didn't mean to approve it prematurely. I can't dismiss my review since it's still in draft form 😅 .

src/data/thunks.js
```diff
@@ -98,6 +97,32 @@ export function getProctoringSettings() {
...
+    //   const examAccessTicket = examAccessTicketResponse.cookies.exam_access_ticket;
+    //   // todo: error if access ticket is expired. Is it possible for it to exist if expired?
+    //   // todo: should exam access ticket be set under exam state in slice?
+    //   dispatch(setExamAccessTicket({ examAccessTicket }));
```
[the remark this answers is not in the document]
Zacharis278: I don't think we need this _yet_ based on my above comment on cookies. Unfortunately, we are required move away from using the cookie based on discussions in https://github.com/openedx/edx-platform/pull/31653 but that's another ticket entirely. Let's stick to the original approach here just to get the fetch part working.

src/data/thunks.js
alangsto: I agree that `updateAttemptAfter` is not the best place for this logic to be included. That function is mean to update the state of an attempt, and while the access token logic is related, I wouldn't necessarily say it's tied to the state of an attempt. 

Additionally, `updateAttemptAfter` would be called on every ping (a regular timed interval), which contrasts the original decision to fetch an access token each time new content should be rendered.

Zacharis278: this makes sense to me as standalone since the trigger for this isn't tied to any of the existing thunks as Alie mentions. We need to kick off this function every time the unit inside xblock iframe url changes. I think you'll have to do some digging around frontend-app-learning to find a good way to accomplish that since as of now this library doesn't know when that changes.

src/data/thunks.js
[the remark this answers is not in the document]
varshamenon4: Ah thanks for the context here.

src/api.js
```diff
@@ -0,0 +1,6 @@
+import { getExamAccessToken, store } from './data';
+
+export default async function checkExamAccessToken() {
```
varshamenon4, line 3: This is the call that will happen in the Unit. Frontend-app-learning is where I'll check that the unit is part of an exam and then update the url to include the access token.

src/data/thunks.js
```diff
@@ -98,6 +99,27 @@ export function getProctoringSettings() {
   };
 }
 
+export function getExamAccessToken() {
+  return async (dispatch, getState) => {
+    const { exam } = getState().examState;
+    if (!exam.id) {
+      logError('Failed to get exam access token. No exam id.');
+      handleAPIError(
+        // todo: check what user error is here, change error?
```
varshamenon4, line 108: Question: should there be a more clear error?

src/data/thunks.js
```diff
@@ -98,6 +99,27 @@ export function getProctoringSettings() {
...
+        { message: 'Failed to fetch exam access token. No exam id was found.' },
+        dispatch,
+      );
+      return;
+    }
+    try {
+      const examAccessToken = await fetchExamAccessToken(exam.id);
+      dispatch(setExamAccessToken({ examAccessToken }));
```
varshamenon4, line 116: I am saving the exam access token to the exam state so I can retrieve from the unit level.

varshamenon4 (commented):
The logic in frontend-lib-special-exams should be complete here. The next PR will be in frontend-app-learning where the call will be made and the iframe url is updated with the token.

src/index.jsx
```diff
@@ -1,2 +1,4 @@
 export { default } from './core/SequenceExamWrapper';
 export { default as OuterExamTimer } from './core/OuterExamTimer';
+export { default as checkExamAccessToken } from './api';
+export { store } from './data';
```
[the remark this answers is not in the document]
varshamenon4: That makes sense. Yes, I could use the checkExamAccessToken to fetch/return the access token response. The reason I had exposed the library's store here was in order to check whether this Unit was part of the exam, so I would only make the call to get the access token if it was an exam. So that would maybe be the tradeoff of not exposing the store. Since the frontend-app-learning store doesn't include the exam state I believe. So I think these are the options:

- check exam state from unit before making call to get exam access token (would require store from lib)
- check if exam by looking at content/course id. This would require adding the content id (aka sequence id) as a prop in the unit and then calling to exams to check if it corresponds to an exam. (this call could be another api function from frontend lib special exams maybe)
- always call checkExamAccessToken from the Unit and have this api function return the token if there is one.

src/data/thunks.js
varshamenon4: That's fair! Sounds good. Given this and the conversation about potentially just having an api function that returns the exam access token, I'm wondering if we should remove all of the logic from thunks and simply do the logic in the checkExamAccessToken function. Since we don't need to store the api error/exam access token (since we are just returning this), I think a thunk is probably not necessary. Since thunks typically just update the state and don't actually return anything? Or rather they technically return a function that updates state. 

What do you think?

src/index.jsx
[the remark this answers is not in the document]
Zacharis278: even if we're always calling `checkExamAccessToken` from the unit, `checkExamAccessToken` can decide if it's an exam and weather or not to make the API call correct?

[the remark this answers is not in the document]
Zacharis278: also I think I was a bit hasty with suggesting checkAccessToken both check AND return since that probably it may not be a very react friendly way to do things. I think that depends on how to best deal with making for the xblock doesn't render until that info is available.

Zacharis278 (commented):
approach looks good. I think we're missing a chunk of test coverage though. Are you still working on this?

src/api.js
```diff
@@ -0,0 +1,16 @@
+import { getExamAccessToken, store } from './data';
+
+export function isExam() {
```
Zacharis278, line 3: suggestion based on our slack thread. It seems like we're still going to end up calling `fetchExamAccessToken` even if EXAMS_BASE_URL is not set for the current environment. Maybe this function should actually be `examRequiresAccessToken()` and check for EXAMS_BASE url in addition to the exam id so we can prevent that from happening?

varshamenon4: Ah good idea! yeah, I was just trying to prevent the error from surfacing, but this is a better approach. I'll update that and testing now.

varshamenon4: Updated!

Zacharis278 (approved):
looking good! 🚀

PR #87, opened 2023-03-08
feat: call edx-exams if exams base url is defined

alangsto:
## [MST-1643](https://2u-internal.atlassian.net/browse/MST-1643)

If the exams service is configured, we should call `edx-exams` endpoints instead of `edx-proctoring`.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,2 +1,3 @@
 LANGUAGE_PREFERENCE_COOKIE_NAME='openedx-language-preference'
 LMS_BASE_URL='http://localhost:18000'
+EXAMS_BASE_URL='http://localhost:18740'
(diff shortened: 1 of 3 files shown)


alangsto:
Wanted to note that I've intentionally left two proctoring endpoints untouched, as those will be updated as part of https://2u-internal.atlassian.net/browse/MST-1797 and https://2u-internal.atlassian.net/browse/MST-1792

.env.test
```diff
@@ -1,2 +1,3 @@
 LANGUAGE_PREFERENCE_COOKIE_NAME='openedx-language-preference'
 LMS_BASE_URL='http://localhost:18000'
+EXAMS_BASE_URL='http://exams'
```
varshamenon4, line 3: Question: is there a reason this can be 'http://exams' and not 'http://localhost:18740', or is this moot since it's just for testing and doesn't really matter since it's being mocked anyway?

varshamenon4 (approved):
Looks great! Just had one question but not blocking approval! Approved :)

PR #88, opened 2023-03-14
feat: separate fetch exam attempt endpoint

alangsto:
## [MST-1797](https://2u-internal.atlassian.net/browse/MST-1797)

Because the data expected by `fetchExamAttemptsData` is now split up into two separate endpoints, ensure that the frontend calls both and combines that data. 

This PR also fixes a bug in the URL path that the frontend tries to hit that I noticed during testing.

Steps for testing with edx-proctoring:
* Created timed exam
* Verified that timed exam was created via LMS admin
* As a verified learner, tested that exam timer was visible on course home page and in exam


src/data/api.js
```diff
@@ -5,14 +5,26 @@ import { ExamAction } from '../constants';
...
+      `${getConfig().LMS_BASE_URL}${BASE_API_URL}/course_id/${courseId}`,
+    );
+    if (sequenceId) {
+      url.searchParams.append('content_id', sequenceId);
+    }
+    url.searchParams.append('is_learning_mfe', true);
+    const urlResponse = await getAuthenticatedHttpClient().get(url.href);
+    data = urlResponse.data;
+  } else {
+    const examUrl = new URL(`${getConfig().EXAMS_BASE_URL}/exam/attempt/course_id/${courseId}/content_id/${sequenceId}`);
```
ilee2u, line 20: NIT: I personally would group these by the data type, e.g:

- examURL
- examResponse
- data = examResponse.data

- activeAttemptUrl
- activeAttemptResponse
- data.active_attempt = ...

This is very minor thing though

src/data/api.js
```diff
@@ -5,14 +5,26 @@ import { ExamAction } from '../constants';
...
+  if (!getConfig().EXAMS_BASE_URL) {
+    const url = new URL(
+      `${getConfig().LMS_BASE_URL}${BASE_API_URL}/course_id/${courseId}`,
+    );
+    if (sequenceId) {
+      url.searchParams.append('content_id', sequenceId);
+    }
+    url.searchParams.append('is_learning_mfe', true);
+    const urlResponse = await getAuthenticatedHttpClient().get(url.href);
+    data = urlResponse.data;
```
alangsto, line 18: I had to modify this to no longer unpack the data value because of how that variable was scoped. In order to not return within each conditional, I defined the variable prior to the conditional, modify within each branch, and then return after

src/data/redux.test.jsx
```diff
@@ -665,13 +668,32 @@ describe('Data layer integration tests', () => {
...
+      axiosMock.onGet(activeAttemptURL).reply(200, { attempt });
+
+      await executeThunk(thunks.getExamAttemptsData(courseId, contentId), store.dispatch);
+
+      expect(axiosMock.history.get[0].url).toEqual(examURL);
+      expect(axiosMock.history.get[1].url).toEqual(activeAttemptURL);
+
+      const state = store.getState();
+      expect(state).toMatchSnapshot();
+    });
```
ilee2u, line 697: Mostly a clarifying question for myself here, it seems like the edx-proctoring functionality/tests remain the same, and you're just adding in tests for edx-exams?

ilee2u (commented):
Not an expert on everything going on here, but I'd say it looks pretty good from what I can tell. Just left a couple notes/questions here and there.

src/data/api.js
alangsto: I was grouping by order (i.e. create the URL for each, call each endpoint, then create the data response), but I can see how either are good

src/data/redux.test.jsx
alangsto: Yep! edx-proctoring behavior is already covered by the existing tests, so I am just adding a test case to cover the behavior for if edx-exams is enabled.

src/data/api.js
```diff
@@ -5,14 +5,26 @@ import { ExamAction } from '../constants';
 const BASE_API_URL = '/api/edx_proctoring/v1/proctored_exam/attempt';
 
 export async function fetchExamAttemptsData(courseId, sequenceId) {
```
ilee2u, line 7: Is there a reason that we don't have multi-line comments describing what each of these functions do as we do in the back-end?

alangsto: We typically don't leave comments for the api functions because they are essentially wrappers - they call an endpoint and return the data. Here's an example of a more mature api.js file: https://github.com/openedx/frontend-app-learning/blob/master/src/course-home/data/api.js

src/data/api.js
```diff
@@ -4,15 +4,52 @@ import { ExamAction } from '../constants';
...
-    `${getConfig().LMS_BASE_URL}${BASE_API_URL}/course_id/${courseId}`,
-  );
-  if (sequenceId) {
-    url.searchParams.append('content_id', sequenceId);
+  let data;
+  if (!getConfig().EXAMS_BASE_URL) {
+    const url = new URL(
+      `${getConfig().LMS_BASE_URL}${BASE_API_URL}/course_id/${courseId}`,
+    );
+    if (sequenceId) {
```
Zacharis278, line 19: this should always be called with a sequence id now right?

src/data/api.js
```diff
@@ -4,15 +4,52 @@ import { ExamAction } from '../constants';
...
+}
+
+//
+export async function fetchLatestAttempt(courseId) {
+  let data;
+  if (!getConfig().EXAMS_BASE_URL) {
+    const url = new URL(
+      `${getConfig().LMS_BASE_URL}${BASE_API_URL}/course_id/${courseId}`,
+    );
+    url.searchParams.append('is_learning_mfe', true);
```
Zacharis278, line 43: 🤔 wonder if we can drop setting this everywhere at this point, there's no case where this isn't true anymore.

Zacharis278 (approved):
a few small questions but looks good 👍

src/data/api.js
alangsto: Is Open edX only using the learning MFE for exams now?

src/data/api.js
alangsto: Let me verify, but I'm pretty sure

alangsto: Yeah, it should always have a sequence ID now. Will remove

src/data/api.js
Zacharis278: yeah, by extension of the non-mfe courseware views being deprecated the old exams views are as well.

alangsto: Can create a follow up ticket to change the backend behavior

Zacharis278: @alangsto I already tossed this in https://2u-internal.atlassian.net/browse/MST-1830

PR #92, opened 2023-03-16
Ilee2u/refresh last active attempt

ilee2u:
PR for [MST-1792](https://2u-internal.atlassian.net/jira/software/projects/MST/boards/584?selectedIssue=MST-1792).

Refresh last active attempt via the pollAttempt thunk in the pollExamAttempt api function.

The function has been split such that, if one is only using edx-proctoring, then only a custom URL/endpoint will be called at the edx-proctoring service. Otherwise, if edx-exams is being used in any capacity, then the discretely defined endpoint for LatestExamAttemptView will be called.

Note: This branch is based off of [PR#88](https://github.com/openedx/frontend-lib-special-exams/pull/88)


ilee2u:
Note: it seems like the codecov for the tests is off, though the code that isn't covered seems like it wasn't covered before I worked on this PR. Wondering if this is something that I need to address.

src/data/redux.test.jsx
```diff
@@ -664,14 +679,61 @@ describe('Data layer integration tests', () => {
...
+    it('Should call the exams service to poll/refresh latest attempt data', async () => {
+      // Mock the endpoints that will be called on edx-exams
+      const examURL = `${getConfig().EXAMS_BASE_URL}/api/v1/student/exam/attempt/course_id/${courseId}/content_id/${contentId}`;
+      const activeAttemptURL = `${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`;
+
+      axiosMock.onGet(examURL).reply(200, { exam });
+      axiosMock.onGet(activeAttemptURL).reply(200, { attempt });
+
+      // make sure that the state is initialized by calling "getExamAttemptsData" first
+      await executeThunk(thunks.getExamAttemptsData(courseId, contentId), store.dispatch);
```
alangsto, line 720: getExamAttemptsData is also calling the `activeAttemptUrl`, so I would suggest mocking out two calls to that endpoint: one for the initial state, one to check if things have been updated. That way you can verify that the data has been updated in the snapshot correctly

ilee2u: Hi there,

I figured out how to implement what you said, and ran into a couple of hiccups.

When only executing the `pollAttempt` thunk, the test ends up saving a snapshot of an exam/attempt object with a bunch of empty and null fields, with the apiErrorMsg field set to `Cannot set properties of undefined (setting 'active_attempt')`.

From what I can tell, this may be due to this line of code at the very start of the `pollAttempt` thunk:
```const currentAttempt = getState().examState.activeAttempt;```
If my understanding is correct, the `pollAttempt` thunk requires the active attempt in the UI's state to be initialized before being called.

As such, in my latest commit, the test calls the `getLatestAttemptData` to initialize the data. The attempt that is returned does then have some data filled out, though it still does contain the same apiErrorMsg, alongside some empty fields.

Comment (alangsto) [reply src/data/redux.test.jsx]: Yeah that makes sense, so the state needs to be initialized prior to polling the attempt. 

If I'm understanding correctly, your current test has `getLatestAttemptData` and `pollAttempt` returning the same data (via the URL mocks). Because of that, it's not clear in the test if the data is actually being updated as expected via the poll attempt URL. 

It's possible to return two separate data responses for the same URL (if you have two different mocks, there are a few examples of that in this testing file). That way you can verify via the snapshot that the poll attempt URL is updating the state. 

Let me know if I'm missing something in my understanding though!

File: src/data/redux.test.jsx [hunk 2]
```diff
@@ -708,13 +708,17 @@ describe('Data layer integration tests', () => {
       expect(state).toMatchSnapshot();
     });
 
-    it('Should call the exams service to get latest attempt data', async () => {
-      const attemptDataUrl = `${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`;
-      axiosMock.onGet(attemptDataUrl).reply(200, { attempt });
+    it('Should call the exams service to poll/refresh latest attempt data', async () => {
```
Zacharis278, line 711: NIT: I think the original language here is more accurate since the polling part happens outside this function and this really does just 'get the latest attempt'

Zacharis278 (commented):
Wow this is actually pretty clean I thought we would have to make more changes to get this working 🎉 .

@ilee2u were you able to manually test this by starting an exam and watching the polling behavior function?

src/data/redux.test.jsx
alangsto: If you're not calling `getLatestAttemptsData`, can you restore the original test then? Just to make sure that the thunk is being tested.

ilee2u:
> Wow this is actually pretty clean I thought we would have to make more changes to get this working 🎉 .
> 
> @ilee2u were you able to manually test this by starting an exam and watching the polling behavior function?

I'm actually not too sure how to test this manually. Should I run the app in my browser or is there a way I can manually execute these commands with some dummy data?

src/data/thunks.js
```diff
@@ -237,7 +237,7 @@ export function pollAttempt(url) {
         ...currentAttempt,
         time_remaining_seconds: data.time_remaining_seconds,
         accessibility_time_string: data.accessibility_time_string,
-        attempt_status: data.status,
+        attempt_status: data.attempt_status,
```
alangsto, line 240: this change isn't going to be backwards compatible with edx-proctoring. If you need to update the dictionary key, I would do that in api.js where you know if you are calling edx-exams

src/data/__snapshots__/redux.test.jsx.snap
```diff
@@ -4,21 +4,23 @@ exports[`Data layer integration tests Test exams IDA url Should call the exams s
...
-      "exam_display_name": "timed",
-      "exam_started_poll_url": "/api/edx_proctoring/v1/proctored_exam/attempt/1",
-      "exam_type": "a timed exam",
-      "exam_url_path": "http://localhost:2000/course/course-v1:test+special+exam/block-v1:test+special+exam+type@sequential+block@abc123",
-      "in_timed_exam": true,
-      "software_download_url": "",
-      "taking_as_proctored": false,
-      "time_remaining_seconds": 1799.9,
-      "total_time": "30 minutes",
+      "attempt": Object {
```
alangsto, line 7: This shouldn't be a nested dictionary, the snapshot for this test was correct before. If you look at the other existing tests, you'll see that the right shape is {activeAttempt: ...}, not {activeAttempt: {attempt: ...}}

src/data/api.js
```diff
@@ -7,7 +7,7 @@ const BASE_API_URL = '/api/edx_proctoring/v1/proctored_exam/attempt';
 async function fetchActiveAttempt() {
   const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
   const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
-  return activeAttemptResponse.data.attempt;
+  return activeAttemptResponse.data;
```
alangsto, line 10: Thanks for catching this, the backend is only returning the attempt data as is

src/data/api.js
```diff
@@ -52,7 +51,21 @@ export async function fetchLatestAttempt(courseId) {
...
+      `${getConfig().LMS_BASE_URL}${url}`,
+    );
+    const urlResponse = await getAuthenticatedHttpClient().get(edxProctoringURL.href);
+    data = urlResponse.data;
+  } else {
+    data = await fetchActiveAttempt();
+    data = data.attempt;
+
+    // Update dictionary to have correct status key for legacy compatibility
+    Object.defineProperty(data, 'status', Object.getOwnPropertyDescriptor(data, 'attempt_status'));
```
alangsto, line 66: the data is just a dictionary, so does it need to be accessed like this? Or could you just say `data.status = data.attempt_status`? 

Also how does this work when edx-exams passes the request to edx-proctoring? edx-proctoring will still still return `status`, so it is probably worth checking if that key already exists in the data.

alangsto (changes requested):
I think there's still some mismatch with the shape of the data being returned

src/data/api.js
ilee2u: Good point. I'm going to make that change now.

ilee2u:
Ok, I fixed the data shape and tests, and fixed the status/attempt_status key piece in pollExamAttempt()

alangsto (approved):
LGTM 👍

PR #94, opened 2023-03-24
feat: refresh accessibility string in UI

ilee2u:
PR for [MST-1799](https://2u-internal.atlassian.net/browse/MST-1799)

Copied over logic from edx-proctoring into this MFE to generate accessibility_time_string on the frontend.


ilee2u:
NOTES:
- I literally just copied over the "humanized_time" function from edx-proctoring and called it within another helper function that actually generates the string. This was more to just get functionality working. I'm totally up to refactoring this if need be.
- All of the tests passed, though I'm unsure if I have to add any new ones.

src/data/thunks.js
```diff
@@ -215,6 +215,63 @@ export function skipProctoringExam() {
...
 
+/*
+Converts the given value in minutes to a more human readable format
+1 -> 1 Minute
+2 -> 2 Minutes
+60 -> 1 hour
+90 -> 1 hour and 30 Minutes
+120 -> 2 hours
+*/
+function humanizedTime(timeInMinutes) {
```
alangsto, line 226: You are passing in seconds, but the function is expecting the time to be passed in in minutes. Should be easy enough to convert from seconds to minutes in the function!

alangsto: wait never mind, just saw the other function sorry!

src/data/thunks.js
```diff
@@ -235,8 +292,8 @@ export function pollAttempt(url) {
       const data = await pollExamAttempt(url);
       const updatedAttempt = {
         ...currentAttempt,
-        time_remaining_seconds: data.time_remaining_seconds,
-        accessibility_time_string: data.accessibility_time_string,
+        timeRemainingSeconds: data.timeRemainingSeconds,
```
alangsto, line 295: I think this should still be in snake case, because that's the field being returned by the backend

src/data/thunks.js
alangsto: On another note though, I don't think this logic belongs in thunks.js

ilee2u:
NOTES:

- Moved accessibility string logic to api.js in a helper function
- Restored backend variable names to snake case to match
- Streamlined the logic in the `humanizedTime()` function
- Wondering if I ought to make separate tests for the `humanizedTime()` function to test different inputs/outputs that aren't currently tested in redux.test.jsx (and furthermore, whether I should make such tests in that file or a different one)

ilee2u:
NOTES:

- Moved accessibility string logic to api.js in a helper function
- Restored backend variable names to snake case to match
- Streamlined the logic in the `humanizedTime()` function
- Wondering if I ought to make separate tests for the `humanizedTime()` function to test different inputs/outputs that aren't currently tested in redux.test.jsx (and furthermore, whether I should make such tests in that file or a different one)

ilee2u:
NOTES:

- Moved accessibility string logic to api.js in a helper function
- Restored backend variable names to snake case to match
- Streamlined the logic in the `humanizedTime()` function
- Wondering if I ought to make separate tests for the `humanizedTime()` function to test different inputs/outputs that aren't currently tested in redux.test.jsx (and furthermore, whether I should make such tests in that file or a different one)

src/data/redux.test.jsx
```diff
@@ -736,4 +736,6 @@ describe('Data layer integration tests', () => {
       expect(beforeState).not.toEqual(afterState); // Test that the state was updated when polled
     });
   });
+  // TODO: Insert tests somewhere in here that just test `generateAccessibilityString()` to make sure it works right.
+  // TODO: Also, move that function to the api.js? maybe??? idk.
```
alangsto, line 740: I think it would be worth it to add some additional tests here for generateAccessibilityString, especially to increase code coverage

Zacharis278:
I was intending to comment that we need to consider internationalization/translations for these strings but I noticed the backend today also doesn't support this. I guess for a screen reader it muse be ok or even desirable to always be english? To avoid future confusion I'd recommend we leave a comment explaining why these strings are not translated.

Zacharis278:
@ilee2u your findings on how this timer can be 'out of sync' brings up a few points.

The way this works is the screen reader alerts the learner each time the underlying value changes. As a result, if the time is 'stale' it isn't so much of an issue. Changing the polling interval would actually result in the page interrupting the learner every 30s which seems aggressive. TBH, even every 1 minute as we have it today doesn't seem ideal.

Since this value isn't provided by the API anymore it doesn't seem to me that this should be tied to pollExam at all. I think we could move this logic entirely within the UI component and remove `accessibility_time_string` from state. That way we can update the rendered value here as frequently or infrequently as we choose without creating additional API requests (and increasing load on our API as a result). I know this is a pretty big deviation from what you have written here but does that make sense?

ilee2u:
> @ilee2u your findings on how this timer can be 'out of sync' brings up a few points.
> 
> The way this works is the screen reader alerts the learner each time the underlying value changes. As a result, if the time is 'stale' it isn't so much of an issue. Changing the polling interval would actually result in the page interrupting the learner every 30s which seems aggressive. TBH, even every 1 minute as we have it today doesn't seem ideal.
> 
> Since this value isn't provided by the API anymore it doesn't seem to me that this should be tied to pollExam at all. I think we could move this logic entirely within the UI component and remove `accessibility_time_string` from state. This is ultimately just driven by `time_remaining_seconds` anyway. That way we can update the rendered value here as frequently or infrequently as we choose without creating additional API requests (and increasing load on our API as a result). I know this is a pretty big deviation from what you have written here but does that make sense?

Makes sense, I made changes to only generate the `accessibility_time_string` for its respective <span> in the UI.

ilee2u:
Also, I smoke tested this with a few different hour/minute time limits and everything seems to work fine.

src/timer/TimerProvider.jsx
```diff
@@ -53,12 +52,6 @@ const TimerServiceProvider = ({
     },
   ).join(':');
 
-  const pollExam = () => {
```
Zacharis278, line 56: the changes to this file seem to be removing the polling with the backend entirely, is that intended? We still need this to make sure the JS timer stays in sync with the backend

src/timer/CountDownTimer.jsx
```diff
@@ -9,15 +9,44 @@ import { TimerContext } from './TimerProvider';
  */
 const CountDownTimer = injectIntl((props) => {
   const timer = useContext(TimerContext);
+  const timeString = timer.getTimeString();
   const [isShowTimer, showTimer, hideTimer] = useToggle(true);
   const { intl } = props;
 
+  const generateAccessbilityString = () => {
+    const hours = parseInt(timeString.substring(0, 2), 10);
```
Zacharis278, line 17: We should find a way not to use another "display string" as the source of truth for this. Especially since we're parsing _back_ into an int value. Maybe we can use the timeState object from the TimerProvider or create some other property for this?

Also I'm not seeing how this string would update, it seems to me it would get stuck at whatever timeString is the first time this component renders?

src/data/api.js
```diff
@@ -51,23 +51,23 @@ export async function fetchLatestAttempt(courseId) {
 }
 
 export async function pollExamAttempt(url) {
-  let data;
+  let activeAttemptResponse;
```
Zacharis278, line 54: why did we need to change this file?

Zacharis278 (changes requested):

src/timer/CountDownTimer.jsx
```diff
@@ -9,15 +9,44 @@ import { TimerContext } from './TimerProvider';
  */
 const CountDownTimer = injectIntl((props) => {
   const timer = useContext(TimerContext);
+  const timeString = timer.getTimeString();
   const [isShowTimer, showTimer, hideTimer] = useToggle(true);
   const { intl } = props;
 
+  const generateAccessbilityString = (timeState) => {
+    const { hours } = timeState;
```
Zacharis278, line 17:
```suggestion
    const { hours, minutes } = timeState;
```
tip: you can do this in one line if you want

Zacharis278 (commented):
Looks good! Could you add some tests for the a11y string to the CountDownTimer?

src/timer/CountDownTimer.test.jsx
```diff
@@ -293,4 +293,58 @@ describe('ExamTimerBlock', () => {
...
+    '2 hours and 29 minutes': 9000,
+    '1 hour and 29 minutes': 5400,
+    '2 hours and 1 minute': 7320,
+    '1 hour and 1 minute': 3720,
+    '2 hours and 0 minutes': 7260,
+    '1 hour and 0 minutes': 3660,
+    '29 minutes': 1800,
+  };
+  Object.keys(timesToTest).forEach((time) => {
+    it('Accessibility time string appears as expected based on time in seconds remaining', async () => {
```
Zacharis278, line 308:
```suggestion
    it( `Accessibility time string appears as expected based on ${timesToTest[time]} seconds remaining`, async () => {
```
small suggestion, this is going to output a bunch of tests with the exact same name "ExamTimerBlock > Accessibility time string appears as expected based on time in seconds remaining" so if one fails you can't tell what test it is. I think you could add the `time` variable into the name of the test.

src/data/slice.js
```diff
@@ -51,9 +51,10 @@ export const examSlice = createSlice({
         exam_display_name: '',
         exam_url_path: '',
         time_remaining_seconds: null,
+        low_threshold_sec: null,
```
Zacharis278, line 54: these values don't exist in state anymore so I think this is a merge issue

Zacharis278 (approved):
two small suggestions but otherwise this looks good

PR #95, opened 2023-04-05
feat: remove use of proctoring settings and use config values

alangsto:
## [MST-1833](https://2u-internal.atlassian.net/browse/MST-1833)

Several of the proctoring settings used for displaying information on various proctoring interstitials should be set as frontend config values to avoid making unnecessary calls to the backend, and because the frontend can now be aware of these values. 

This PR replaces `platform_name`, `contact_us`, and `link_urls` with pre-existing or newly created config variables. Several of the variables contained in the proctoringSettings were not even being used, so I also removed those.


src/instructions/proctored_exam/prerequisites-instructions/index.jsx
```diff
@@ -9,17 +9,15 @@ import Footer from '../Footer';
 
 const PrerequisitesProctoredExamInstructions = ({ skipProctoredExam }) => {
   const state = useContext(ExamStateContext);
-  const { exam, proctoringSettings, allowProctoringOptOut } = state;
+  const { exam, allowProctoringOptOut } = state;
   const { prerequisite_status: prerequisitesData } = exam;
   const { pending_prerequisites: pending, failed_prerequisites: failed } = prerequisitesData;
-  const { platform_name: platformName } = proctoringSettings;
```
alangsto, line 15: can remove this completely because it was being passed in as a prop to `FailedPrerequisitesProctoredExamInstructions`

src/data/__factories__/proctoringSettings.factory.js
```diff
@@ -2,14 +2,6 @@ import { Factory } from 'rosie'; // eslint-disable-line import/no-extraneous-dep
 
 Factory.define('proctoringSettings')
   .attrs({
-    platform_name: 'Your Platform Name Here',
-    contact_us: 'info@example.com',
-    link_urls: {
-      contact_us: 'https://example.com/contact_us/',
-      faq: 'https://example.com/faq/',
-      online_proctoring_rules: 'https://example.com/online_proctoring_rules/',
-      tech_requirements: 'https://example.com/tech_requirements/',
```
alangsto, line 11: couldn't find any uses of `tech_requirements` in the frontend lib so I opted to remove it

src/exam/Exam.jsx
```diff
@@ -57,10 +57,12 @@ const Exam = ({
   const proctoredExamTypes = [ExamType.ONBOARDING, ExamType.PRACTICE, ExamType.PROCTORED];
 
   useEffect(() => {
-    if (examId) {
-      getProctoringSettings();
-    }
     if (proctoredExamTypes.includes(examType)) {
+      // only fetch proctoring settings for a proctored exam
+      if (examId) {
```
alangsto, line 62: moved this check inside to avoid calling the proctoring settings endpoint for timed exams (as it's no longer needed)

Zacharis278 (approved):
hurray for removing lines of code 👍

PR #96, opened 2023-04-06
route attempt endpoints to correct backend

Zacharis278:
### [MST-1831](https://2u-internal.atlassian.net/browse/MST-1831)

Whenever an exam attempt endpoint is hit (create/update) this will first check if that exam has a property indicating it is managed by edx-proctoring instead of edx-exams. If so, route the request there.

When writing out tests for this I noticed that almost none of our test cases are running against a version of the app with EXAMS_BASE_URL set. Since that's the direction we're headed I've made that the default and added an extra test case to each thunk to validate the old integration is still working. I fixed a few issues with the existing tests but there's still work that could be done. This took a ton of lines because of how these tests are structured. A bigger refactor could eliminate o lot code in here but I didn't want to re-invent too many things at once so I stuck with the existing structure as best I could.

### DIFF
--- a/src/data/__snapshots__/redux.test.jsx.snap
+++ b/src/data/__snapshots__/redux.test.jsx.snap
@@ -7,134 +7,6 @@ Object {
 }
 `;
 
-exports[`Data layer integration tests Test exams IDA url Should call the exams service to fetch attempt data 1`] = `
-Object {
-  "examState": Object {
-    "activeAttempt": Object {
-      "attempt_code": "",
-      "attempt_id": 1,
-      "attempt_status": "started",
-      "course_id": "course-v1:test+special+exam",
-      "desktop_application_js_url": "",
-      "exam_display_name": "timed",
-      "exam_started_poll_url": "/api/edx_proctoring/v1/proctored_exam/attempt/1",
-      "exam_type": "a timed exam",
-      "exam_url_path": "http://localhost:2000/course/course-v1:test+special+exam/block-v1:test+special+exam+type@sequential+block@abc123",
-      "in_timed_exam": true,
-      "software_download_url": "",
-      "taking_as_proctored": false,
-      "time_remaining_seconds": 1799.9,
-      "total_time": "30 minutes",
-    },
-    "allowProctoringOptOut": false,
-    "apiErrorMsg": "",
-    "exam": Object {
-      "attempt": Object {
-        "attempt_code": "",
-        "attempt_id": 1,
-        "attempt_status": "started",
-        "course_id": "course-v1:test+special+exam",
-        "desktop_application_js_url": "",
-        "exam_display_name": "timed",
-        "exam_started_poll_url": "/api/edx_proctoring/v1/proctored_exam/attempt/1",
-        "exam_type": "a timed exam",
-        "exam_url_path": "http://localhost:2000/course/course-v1:test+special+exam/block-v1:test+special+exam+type@sequential+block@abc123",
-        "in_timed_exam": true,
-        "software_download_url": "",
-        "taking_as_proctored": false,
-        "time_remaining_seconds": 1799.9,
-        "total_time": "30 minutes",
-      },
-      "backend": "test",
-      "content_id": "block-v1:test+special+exam+type@sequential+block@abc123",
-      "course_id": "course-v1:test+special+exam",
-      "due_date": null,
-      "exam_name": "prock exam",
-      "external_id": null,
-      "hide_after_due": false,
-      "id": 1,
-      "is_active": true,
-      "is_practice_exam": false,
-      "is_proctored": false,
-      "prerequisite_status": Object {
-        "are_prerequisites_satisifed": true,
-        "declined_prerequisites": Array [],
-        "failed_prerequisites": Array [],
-        "pending_prerequisites": Array [],
-        "satisfied_prerequisites": Array [],
-      },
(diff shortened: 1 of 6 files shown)


src/data/thunks.js
```diff
@@ -152,7 +152,7 @@ export function startTimedExam() {
       return;
     }
     await updateAttemptAfter(
-      exam.course_id, exam.content_id, createExamAttempt(exam.id),
+      exam.course_id, exam.content_id, createExamAttempt(exam.id, exam.use_legacy_attempt_api),
```
Zacharis278, line 155: we could just pass the whole exam in here instead of creating an new argument. I chose not to do this because it's an untyped dictionary and I dislike passing around a dict of mystery contents instead of explicit args. Also, I'd hope this gets removed with an edx-proctoring DEPR anyway.

src/data/redux.test.jsx
```diff
@@ -185,59 +232,110 @@ describe('Data layer integration tests', () => {
...
+
+        await executeThunk(thunks.getExamAttemptsData(courseId, contentId), store.dispatch);
+        let state = store.getState();
+        expect(state.examState.activeAttempt.attempt_status).toBe(ExamStatus.STARTED);
+
+        await executeThunk(thunks.stopExam(), store.dispatch, store.getState);
+        state = store.getState();
+        expect(state.examState.exam.attempt.attempt_status).toBe(ExamStatus.READY_TO_SUBMIT);
+        expect(axiosMock.history.put[0].url).toEqual(updateAttemptStatusLegacyUrl);
+        expect(axiosMock.history.put[0].data).toEqual(JSON.stringify({ action: 'stop' }));
```
Zacharis278, line 253: added a bunch of checks like this to the original tests. In many cases our assert was just checking that the state is updated which is really only testing that we put the right mock response into the second fetchAttempts call.

src/data/redux.test.jsx
```diff
@@ -250,25 +348,48 @@ describe('Data layer integration tests', () => {
     const readyToSubmitAttempt = Factory.build('attempt', { attempt_status: ExamStatus.READY_TO_SUBMIT });
     const readyToSubmitExam = Factory.build('exam', { attempt: readyToSubmitAttempt });
 
-    it('Should stop exam, and update attempt and exam', async () => {
-      axiosMock.onGet(fetchExamAttemptsDataUrl).replyOnce(200, { exam: readyToSubmitExam, active_attempt: {} });
-      axiosMock.onGet(fetchExamAttemptsDataUrl).reply(200, { exam, active_attempt: attempt });
-      axiosMock.onPost(updateAttemptStatusUrl).reply(200, { exam_attempt_id: attempt.attempt_id });
```
Zacharis278, line 256: in a few cases where the mocked method isn't even correct but because we're never asserting it's called the test was passing.

src/data/redux.test.jsx
```diff
@@ -711,173 +994,16 @@ describe('Data layer integration tests', () => {
...
-      expect(axiosMock.history.get[0].url).toEqual(activeAttemptURL);
-      expect(axiosMock.history.get[1].url).toEqual(activeAttemptURL);
-
-      expect(afterState).toMatchSnapshot();
-      expect(beforeState).not.toEqual(afterState); // Test that the state was updated when polled
-    });
-  });
-});
-
-describe('External API integration tests', () => {
```
Zacharis278, line 823: moved these to a different file since they don't have to do with the redux store and this file is already ginormous

src/data/redux.test.jsx
```diff
@@ -711,173 +994,16 @@ describe('Data layer integration tests', () => {
...
-      await executeThunk(thunks.getExamAttemptsData(courseId, contentId), store.dispatch);
-      await executeThunk(thunks.startTimedExam(), store.dispatch, store.getState);
       await executeThunk(thunks.examRequiresAccessToken(), store.dispatch, store.getState);
 
       const state = store.getState();
       expect(state.examState.examAccessToken.exam_access_token).toBe('');
     });
   });
 
-  describe('Test exams IDA url', () => {
```
Zacharis278, line 733: I think I've rolled everything covered in here up into the above features

src/data/__snapshots__/redux.test.jsx.snap
```diff
@@ -333,12 +224,80 @@ Object {
 }
 `;
 
-exports[`Data layer integration tests Test resetExam Should reset exam, and update attempt and exam 1`] = `
+exports[`Data layer integration tests Test resetExam Should reset exam attempt 1`] = `
 Object {
   "examState": Object {
     "activeAttempt": null,
     "allowProctoringOptOut": false,
-    "apiErrorMsg": "Request failed with status code 404",
```
Zacharis278, line 341: example of a test that was basically just asserting it was throwing error...

alangsto: Are some of these test cases missing now? I agree that this perhaps wasn't the best approach but it looks like there are a few test cases missing for when the API errors now

Zacharis278: yeah I just dropped a few of these because the code  path for an api error is the same for most of these and we're just testing the same thing over and over and over.

That's probably a good test we could isolate and pull out into its own describe block. I'll add a comment in the code to mention this.

alangsto (approved):

PR #97, opened 2023-04-06
feat: add null state check

varshamenon4:
Part of [MST-1599](https://2u-internal.atlassian.net/browse/MST-1599).

Based on feedback from [this PR](https://github.com/openedx/frontend-app-learning/pull/1083) to fetch exam access token, we want to handle the null case when there is no exam. Thus the api functions here should include a null check to handle this.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -2,15 +2,22 @@ import { examRequiresAccessToken, store } from './data';
 
 export function isExam() {
   const { exam } = store.getState().examState;
-  return exam.id !== null;
+  return !!exam?.id;
 }
 
 export function getExamAccess() {
-  const { examAccessToken } = store.getState().examState;
+  const { exam, examAccessToken } = store.getState().examState;
+  if (!exam) {
+    return '';
+  }
   return examAccessToken.exam_access_token;
 }
 
 export async function fetchExamAccess() {
+  const { exam } = store.getState().examState;
   const { dispatch } = store;
+  if (!exam) {
+    return Promise.resolve();
+  }
   return dispatch(examRequiresAccessToken());
 }
--- a/src/api.test.jsx
+++ b/src/api.test.jsx
@@ -1,25 +1,59 @@
+import { Factory } from 'rosie';
+
 import { isExam, getExamAccess, fetchExamAccess } from './api';
+import { store } from './data';
 
 describe('External API integration tests', () => {
-  let store;
+  describe('Test isExam with exam', () => {
+    beforeAll(() => {
+      jest.mock('./data');
+      const mockExam = Factory.build('exam', { attempt: Factory.build('attempt') });
+      const mockToken = Factory.build('examAccessToken');
+      const mockState = { examState: { exam: mockExam, examAccessToken: mockToken } };
+      store.getState = jest.fn().mockReturnValue(mockState);
+    });
 
-  describe('Test isExam', () => {
-    it('Should return false if exam is not set', async () => {
+    afterAll(() => {
+      jest.clearAllMocks();
+      jest.resetAllMocks();
+    });
+
+    it('isExam should return true if exam is set', () => {
+      expect(isExam()).toBe(true);
+    });
+
+    it('getExamAccess should return exam access token if access token', () => {
+      expect(getExamAccess()).toBeTruthy();
+    });
+
+    it('fetchExamAccess should dispatch get exam access token', () => {
+      const dispatchReturn = fetchExamAccess();
+      expect(dispatchReturn).toBeInstanceOf(Promise);
+    });
+  });
+
+  describe('Test isExam without exam', () => {
+    beforeAll(() => {
+      jest.mock('./data');
+      const mockState = { examState: { exam: null, examAccessToken: null } };
+      store.getState = jest.fn().mockReturnValue(mockState);
+    });
+
+    afterAll(() => {
+      jest.clearAllMocks();
+      jest.resetAllMocks();
+    });
+
+    it('isExam should return false if exam is not set', () => {
       expect(isExam()).toBe(false);
     });
-  });
 
-  describe('Test getExamAccess', () => {
-    it('Should return empty string if no access token', async () => {
-      expect(getExamAccess()).toBe('');
+    it('getExamAccess should return empty string if exam access token not set', () => {
+      expect(getExamAccess()).toBeFalsy();
     });
-  });
 
-  describe('Test fetchExamAccess', () => {
-    it('Should dispatch get exam access token', async () => {
-      const mockDispatch = jest.fn(() => store.dispatch);
-      const mockState = jest.fn(() => store.getState);
-      const dispatchReturn = fetchExamAccess(mockDispatch, mockState);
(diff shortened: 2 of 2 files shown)


src/api.js
```diff
@@ -1,16 +1,29 @@
+import { logError } from '@edx/frontend-platform/logging';
 import { examRequiresAccessToken, store } from './data';
 
 export function isExam() {
   const { exam } = store.getState().examState;
+  if (!exam) {
+    return false;
```
alangsto, line 7: How is this check different from what's being returned? If there is no exam set in the state, wouldn't it's value be `null`?

src/api.js
```diff
@@ -1,16 +1,29 @@
...
+  }
   return exam.id !== null;
 }
 
 export function getExamAccess() {
-  const { examAccessToken } = store.getState().examState;
+  const { exam, examAccessToken } = store.getState().examState;
+  if (!exam) {
+    logError('Failed to get exam access token. Not an exam.');
+    return '';
```
alangsto, line 16: you can just use `return;` here!

src/data/redux.test.jsx
```diff
@@ -842,42 +842,10 @@ describe('External API integration tests', () => {
...
-
-  describe('Test isExam', () => {
-    it('Should return false if exam is not set', async () => {
-      expect(isExam()).toBe(false);
-    });
-  });
-
-  describe('Test getExamAccess', () => {
-    it('Should return empty string if no access token', async () => {
-      expect(getExamAccess()).toBe('');
```
alangsto, line 865: It looks like test coverage is now missing for the getExamAccess function, is there a reason these were removed?

varshamenon4: Ah I see! This was actually a duplicate of the existing tests. But I did modify this specific test to handle the rejected promise case, which doesn't call get exam access. So I just need to add that case back. I think one thing that I was having trouble with in the past is that the mock/test store in the tests is different from the store that is checked in the api functions themselves. I was having trouble mocking that but I will look into mocking that now so that I can still make the call with the test store rather than a newly created store.

src/api.js
varshamenon4: My understanding was that exam.id wouldn't be able to be accessed because exam itself is null which is what was leading to the error in the case where the unit wasn't an exam.

Here's the original convo for context: https://github.com/openedx/frontend-app-learning/pull/1083#discussion_r1159956063

alangsto: Ahh got it, I might do something like `exam ? exam.id !== null : false` just to make it more concise

Zacharis278: I'll throw some JS magic in here, the `?.` operator is really useful for null checking a chain of properties. So you can do `exam?.id ? true : false` or even just `!!exam?.id`. `!!` is a common use for 'cast to boolean'.

src/data/redux.test.jsx
varshamenon4: As I was trying to add back test coverage, I realized that store.getState() is really only used in testing and in these API functions. I believe because typically the state is accessed in context within a component but for these API functions, they are happening outside of the component/thunks. 

In the previous PR, I only tested the case where exam id was null, but in order to have coverage of the exam access thunk, I need to reconcile the issue I was having before. Basically, the test store that is initialized in redux doesn't correspond to the store/state that is accessed/called in the api functions (isExam, getExamAccess, fetchExamAccess). When I tried to test, I wasn't sure how to test because the store that the tests are checking are the test store, while the implementation checks the store that is called. So I'm essentially not sure how to mock this out without mocking the entire implementation which doesn't seem correct.

Any ideas, @Zacharis278 or @alangsto ? Thanks!

Zacharis278: I don't think there should be much difference from running the thunk tests since we also aren't initializing any components in that case either. Are you using https://github.com/openedx/frontend-lib-special-exams/blob/3239296249ac225a8eeb8a562e7fc1e06eb9f90c/src/setupTest.js#L48 to setup the store?

src/api.js
```diff
@@ -1,16 +1,34 @@
...
+  }
   return examAccessToken.exam_access_token;
 }
 
 export async function fetchExamAccess() {
-  const { dispatch } = store;
+  console.log("exam state", store.getState().examState);
+  const { dispatch, exam } = store;
+  if (!exam) {
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
varshamenon4: That does make sense to me. The only issue I was running into earlier which is what lead me down this path is, in the case where there is an exam and the access token thunk is dispatched, a promise is returned. So wouldn't a promise also need to be returned in the case where there is no exam? For example, I did get an error when I tried to just `return;` for getExamAccess, since it was inconsistent with the return of an exam access string.

src/api.js
varshamenon4: When I tried to return, I got an error (I believe code quality though I can replicate to see what the error is) that these would result in inconsistent types, since the non-error case returns a string.

src/api.js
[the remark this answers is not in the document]
Zacharis278: the return types should be consistent so I'd agree this should always return a promise. That said, we don't really need the error so a simple `return Promise.resolve()` is likely sufficient.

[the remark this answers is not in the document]
varshamenon4: Okay, I'll tinker with that!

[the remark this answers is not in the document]
varshamenon4: Should I remove the error for getExamAccess above? Since the error isn't really needed? It's just handling the case where there is no exam?

src/api.js
```diff
@@ -2,15 +2,22 @@ import { examRequiresAccessToken, store } from './data';
 
 export function isExam() {
   const { exam } = store.getState().examState;
-  return exam.id !== null;
+  return !!exam?.id;
 }
 
 export function getExamAccess() {
```
varshamenon4, line 8: Note: I removed the logError from both getExamAccess and fetchExamAccess. As we'd discussed, it's not really needed since it's not exactly an error if a unit is not an exam, so we probably don't want extraneous logs.

src/api.js
```diff
@@ -2,15 +2,22 @@ import { examRequiresAccessToken, store } from './data';
...
+    return '';
+  }
   return examAccessToken.exam_access_token;
 }
 
 export async function fetchExamAccess() {
+  const { exam } = store.getState().examState;
   const { dispatch } = store;
+  if (!exam) {
+    return Promise.resolve();
```
varshamenon4, line 20: Note: I had it return a resolved promise. When I was playing with promises, I noticed that even if I did Promise.reject(), I would need to catch the error that is thrown and the promise returned would essentially return a resolved Promise (which was why I had the whole block with the catch error included before). Also if we're not logging an error anyway, then the error part is probably extraneous anyway. Lemme know what you think.

The frontend app learning shouldn't really get here, but even if the exam access token isn't granted, I have a check to only unblock exam access for exams if there is an exam access token (so not allowing for blank exam access tokens).

varshamenon4 (commented):
Ready for re-review. Just added some notes on the changes I made based on feedback.

Zacharis278 (approved):

PR #98, opened 2023-04-10
feat: force version update

Zacharis278:
force a library version update for #96. I merged this w/o a feat tag so the release action didn't run.


ilee2u (approved):
LGTM

PR #100, opened 2023-04-12
feat: update exam total time

alangsto:
## [MST-1836](https://2u-internal.atlassian.net/browse/MST-1836)

With the movement of the "humanized time" logic to the frontend, `edx-exams` will only return the amount of total time left as an integer representing the amount of minutes left in an exam. Because of this change in backend behavior, the frontend now needs to humanize the total time if the data is being returned by `edx-exams`.

In order to accomplish this, I've refactored where the humanized time function is located so that is accessible for both the timer component and the API functions.

### DIFF
--- a/src/instructions/proctored_exam/ReadyToStartProctoredExamInstructions.jsx
+++ b/src/instructions/proctored_exam/ReadyToStartProctoredExamInstructions.jsx
@@ -13,7 +13,7 @@ const ReadyToStartProctoredExamInstructions = () => {
     startProctoredExam,
   } = state;
   const { attempt, reviewPolicy } = exam;
-  const { total_time: examDuration } = attempt;
+  const examDuration = attempt.total_time ? attempt.total_time : exam.total_time;
   const platformName = getConfig().SITE_NAME;
   const rulesUrl = getConfig().PROCTORED_EXAM_RULES_URL;
   const [beginExamClicked, setBeginExamClicked] = useState(false);
--- a/src/timer/CountDownTimer.test.jsx
+++ b/src/timer/CountDownTimer.test.jsx
@@ -296,13 +296,14 @@ describe('ExamTimerBlock', () => {
 
   const timesToTest = {
     // Because times are rounded down, these values are 60 seconds off
-    '2 hours and 29 minutes': 9000,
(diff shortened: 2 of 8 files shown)


src/timer/CountDownTimer.test.jsx
```diff
@@ -296,13 +296,14 @@ describe('ExamTimerBlock', () => {
 
   const timesToTest = {
     // Because times are rounded down, these values are 60 seconds off
```
alangsto, line 298: wasn't sure whether or not to remove this comment yet, wanted to make sure it was okay that these values changed

ilee2u (commented):
- Is StartTimedExamInstructions' examDuration automatically updated correctly?
- Do we have/need tests for Timed/Proctored exam instructions rendering the examDuration?

alangsto:
StartTimedExamInstruction gets `examDuration` from `exam.total_time`, and the value of `total_time` is modified to a humanized time in the api.js file. 

And yes, there is testing for exam duration rendering for a `ready_to_start` exam: https://github.com/openedx/frontend-lib-special-exams/blob/eb0fa82644fd00110776e7c8c0186ef0121eab55/src/instructions/proctored_exam/ProctoredExamInstructions.test.jsx#L84

Probably worth adding a check to some of the other tests that already exist!

src/instructions/proctored_exam/ReadyToStartProctoredExamInstructions.jsx
```diff
@@ -13,7 +13,7 @@ const ReadyToStartProctoredExamInstructions = () => {
     startProctoredExam,
   } = state;
   const { attempt, reviewPolicy } = exam;
-  const { total_time: examDuration } = attempt;
+  const examDuration = attempt.total_time ? attempt.total_time : exam.total_time;
```
alangsto, line 16: I realized that `attempt.total_time` and `exam.total_time` can return different values, as `attempt.total_time` takes resumed attempts into account (meaning the attempt's total time could be less depending on whether or not it is part of a second change attempt). 

I updated this assignment to default to the exam total time in the case where `attempt.total_time` doesn't exist, to solve for edx-exams.

Zacharis278 (approved):
lgtm

PR #107, opened 2023-05-05
fix: missing legacy flag on download attempt call

Zacharis278:
Quick fix to unblock testing in stage. The update call for 'click_download_software' is always hitting edx-exams right now. I didn't replace this with the thunk which is what we probably should be doing long term. The way the thunk is written we'd end up making double calls to our latest attempt endpoint so it requires a bit more  refactoring to get there.


varshamenon4 (approved):
Looks good!

PR #113, opened 2023-08-04
fix: pollAttempt call on download page

Zacharis278:
### [MST-1951](https://2u-internal.atlassian.net/browse/MST-1951)

pair PR with https://github.com/edx/frontend-app-exams-dashboard/pull/11

The network call made by the download interstitial is a bit of a special snowflake it bypasses the thunk and calls the api directly. Added support to the existing 'poll' function to support requesting a specific exam by sequence id in this case.

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -1,14 +1,24 @@
 import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
+import { logError } from '@edx/frontend-platform/logging';
 import { ExamAction } from '../constants';
 import { generateHumanizedTime } from '../helpers';
 
 const BASE_API_URL = '/api/edx_proctoring/v1/proctored_exam/attempt';
 
 async function fetchActiveAttempt() {
+  // fetch 'active' (timer is running) attempt if it exists
   const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
   const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
   return activeAttemptResponse.data;
+}
+
+async function fetchLatestExamAttempt(sequenceId) {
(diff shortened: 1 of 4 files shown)


src/data/api.js
```diff
@@ -1,16 +1,26 @@
...
 const BASE_API_URL = '/api/edx_proctoring/v1/proctored_exam/attempt';
 
 async function fetchActiveAttempt() {
+  // fetch 'active' (timer is running) attempt if it exists
   const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
   const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
   return activeAttemptResponse.data;
 }
 
+async function fetchLatestExamAttempt(sequenceId) {
```
Zacharis278, line 16: basically the same thing as the above function. I broke this out to make reading the behavior a bit easier

src/instructions/proctored_exam/download-instructions/index.jsx
```diff
@@ -50,7 +50,7 @@ const DownloadSoftwareProctoredExamInstructions = ({ intl, skipProctoredExam })
     ? `${getConfig().EXAMS_BASE_URL}/lti/start_proctoring/${attemptId}` : downloadUrl;
 
   const handleDownloadClick = () => {
-    pollExamAttempt(`${pollUrl}?sourceid=instructions`)
+    pollExamAttempt(pollUrl, sequenceId)
```
Zacharis278, line 53: `sourceid=instructions` doesn't do anything. Was just copied behavior from the old frontend.

src/data/api.js
ilee2u: Would it be worth adding a comment noting this so that people don't confused by these functions having similar functionality later on?

ilee2u (approved):
Just a couple of small things, otherwise LGTM

src/data/api.js
Zacharis278: good idea 👍

PR #115, opened 2023-08-25
refactor: updated lock file version check to use new workflow

Mashal-m:
Renamed `lockfileversion-check-v3` to `lockfile-check` in lockfile version file.

# Ticket

[Prototype a way to review NPM lockfiles in PRs](https://github.com/edx/edx-arch-experiments/issues/355)


feanil (approved):
I think this is good to merge given that the coverage issue is independent of the change made here.

PR #116, opened 2023-09-08
feat: LTI proctoring integrity check

Zacharis278:
### [MST-1962](https://2u-internal.atlassian.net/browse/MST-1962)

Covers the case where a user may use their credentials on another browser or machine to load exam problems outside of the proctored app.

- Adds a check to ensure proctoring software is running if page is loaded in state monitored by a tool
- This check runs in the background and does not block page load or user actions but will trigger an exam error if failed.
- As an aside, we need to move everything to a single browser tab flow. I've made that small tweak here.

The implementation of this is specific to the Proctorio integration at the moment. We should consider a more LTI compliant solution if other vendors are to be supported in the future. Or, minimally make the existence of this check configurable. Unfortunately we authorize a user to access content during an exam so there isn't a simple way to ensure that only one client is viewing that page. A solution that works generally would be more complex to introduce.


src/constants.js
```diff
@@ -32,6 +32,7 @@ export const ONBOARDING_ERRORS = [
 export const IS_STARTED_STATUS = (status) => [ExamStatus.STARTED, ExamStatus.READY_TO_SUBMIT].includes(status);
 export const IS_INCOMPLETE_STATUS = (status) => INCOMPLETE_STATUSES.includes(status);
 export const IS_ONBOARDING_ERROR = (status) => ONBOARDING_ERRORS.includes(status);
+export const IS_PROCTORED_STATUS = (status) => IS_STARTED_STATUS(status) || status === ExamStatus.READY_TO_START;
```
alangsto, line 35: could you add a comment explaining the use case for this group of statuses? The statuses are also applicable to timed exams, so I was ~slightly confused reading the name at first (before looking at how it's used in the code)

src/data/redux.test.jsx
```diff
@@ -1095,4 +1095,88 @@ describe('Data layer integration tests', () => {
...
+        removeEventListener: jest.fn(),
+      }));
+    });
+
+    afterEach(() => {
+      jest.clearAllMocks();
+      axiosMock.reset();
+    });
+
+    it('should check application status for LTI proctored exams in a proctored state', async () => {
```
alangsto, line 1118: Not sure if it's overkill, but could a test be added to check that the application status is not checked for non-LTI proctored exams in a proctored state?

alangsto (approved):
One question about an additional test, but otherwise looks good 👍

PR #126, opened 2023-12-12
feat: update checkApp origin requirement

Zacharis278:
The proctorio extension running the exam will use a url on the exams service instead of the MFE url.

More details in https://github.com/edx/edx-exams/pull/227


alangsto (approved):
👍

PR #127, opened 2023-12-20
feat: add support for a technical support URL in the LTI-based provider download instructions

MichaelRoytman:
### Description

**Jira:** [COSMO-130](https://2u-internal.atlassian.net/browse/COSMO-130)
**Dependencies:** https://github.com/edx/edx-exams/pull/231

This commit adds support for displaying a technical support URL for LTI-based providers on the download instructions interstitial. The download instructions will display a technical support URL when it is returned from the proctoring settings backend endpoint. If the technical support URL is not available, then the technical support email and technical support phone number will be used instead.

### Screenshots

**supportURL**

image

**supportEmail, supportPhone**

image

**no supportURL, supportEmail, supportPhone**

image

### DIFF
--- a/src/instructions/proctored_exam/download-instructions/LtiProviderInstructions.jsx
+++ b/src/instructions/proctored_exam/download-instructions/LtiProviderInstructions.jsx
@@ -1,47 +1,70 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { Hyperlink } from '@edx/paragon';
 
 const LtiProviderExamInstructions = ({
-  providerName, supportEmail, supportPhone,
-}) => (
-  <>
-    <p>
-      <FormattedMessage
-        id="exam.DownloadSoftwareProctoredExamInstructions.text1"
-        defaultMessage={'Note: As part of the proctored exam setup, you '
-        + 'will be asked to verify your identity. Before you begin, make '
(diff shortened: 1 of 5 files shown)


src/instructions/proctored_exam/download-instructions/LtiProviderInstructions.jsx
```diff
@@ -1,47 +1,70 @@
...
+  );
+
+  const getSupportText = () => {
+    // We assume that an LTI-based provider will either have a supportURL or a supportEmail and supportPhone.
+    // In the unlikely event a provider has all three, we prioritize the supportURL.
+    if (supportURL) {
+      return (
         <FormattedMessage
-          id="exam.DownloadSoftwareProctoredExamInstructions.supportText"
+          id="exam.DownloadSoftwareProctoredExamInstructions.LTI.supportText.URL"
```
MichaelRoytman, line 21: I noticed that there is a `FormattedMessage` in `RestProviderInstructions.jsx` with the same original ID (`exam.DownloadSoftwareProctoredExamInstructions.supportText`). Does anyone know if that would cause a conflict with this original string? I added `LTI` to differentiate, but I'm not sure if the strings end up namespaced by component or something.

Zacharis278: also not sure but differentiating them seems wise

Zacharis278 (approved):

PR #129, opened 2024-01-05
feat: reduce unnecessary api calls

Zacharis278:
Every sequence navigation renders a new ExamWrapper component causing us to hit the `/latest` endpoint every time. We really only need to do this in a few cases.

1. If the sequence is an exam itself
2. If we've first loaded the UI and there is an active exam in another sequence

If the store already has a latest exam there's no reason to keep checking it on every user navigation, the polling function will already take care of refreshing data. I expect this change to drastically reduce the number of API calls we're generating.

### DIFF
--- a/src/exam/ExamWrapper.test.jsx
+++ b/src/exam/ExamWrapper.test.jsx
@@ -3,7 +3,7 @@ import { Factory } from 'rosie';
 import React from 'react';
 import SequenceExamWrapper from './ExamWrapper';
 import { store, getExamAttemptsData, startTimedExam } from '../data';
-import { render } from '../setupTest';
+import { render, waitFor } from '../setupTest';
 import ExamStateProvider from '../core/ExamStateProvider';
 import { ExamStatus, ExamType } from '../constants';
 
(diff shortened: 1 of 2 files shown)


src/exam/ExamWrapper.test.jsx
```diff
@@ -114,6 +114,48 @@ describe('SequenceExamWrapper', () => {
...
+          <div>children</div>
+        </SequenceExamWrapper>
+      </ExamStateProvider>,
+      { store },
+    );
+
+    // can't figure out this test. For whatever reason the mock getExamAttemptsData
+    // is overriden by the actual value whenever this component is rendered.
+    // This pattern works in other tests.
+    // await waitFor(() => expect(getExamAttemptsData).toHaveBeenCalled());
```
Zacharis278, line 156: this one has me puzzled, I can see the mock is setup properly right before the render call and mocking other values such as isLoading or something custom works. Mocking a thunk gets rewritten somehow during the render/connect process I just can't figure out why.

alangsto (approved):
👍

MichaelRoytman (approved):

PR #131, opened 2024-02-01
Refactor references to examStore to use the useDispatch and useSelector hooks API.

MichaelRoytman:
### Description:

**Jira**: [COSMO-173](https://2u-internal.atlassian.net/browse/COSMO-173)

* feat: rename `examStore` to `specialExams` in preparation for use in `frontend-app-learning`
  * This commit renames the `examStore` to `specialExams`. This is in preparation for the use of the exam reducer in the `frontend-app-learning` React application.
* refactor: replace use of context with thunks and Redux store in instructions components
  * This commit replaces the use of the `ExamStateContext` with the use of thunks and the Redux store in the instructions components.
  * The original pattern was to use the `withExamStore` higher-order component to provide context to the instructions components. This context contained provided the Redux store state and action creators as props by using the `connect` API. This posed a problem for our need to merge the `frontend-app-learning` and `frontend-lib-special-exams` stores, because the special exams store is initialized in this repository and used by the higher-order component. In order to eventually be able to remove the creation of the store in this repository, we have to remove references to the store by interfacing with the Redux more directly by using the `useDispatch` and `useSelector` hooks.
* test: remove references to `ExamStateProvider` from tests for instructions components
  * This commit removes references to the `ExamStateProvider` from tests for instructions components. Because these components have been refactored to no longe rely on the `ExamStateProvider`, they are not required in the component tree.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,12 +1,12 @@
 import { examRequiresAccessToken, store } from './data';
 
 export function isExam() {
-  const { exam } = store.getState().examState;
+  const { exam } = store.getState().specialExams;
   return !!exam?.id;
 }
 
 export function getExamAccess() {
-  const { exam, examAccessToken } = store.getState().examState;
+  const { exam, examAccessToken } = store.getState().specialExams;
   if (!exam) {
     return '';
   }
@@ -14,7 +14,7 @@ export function getExamAccess() {
 }
 
 export async function fetchExamAccess() {
-  const { exam } = store.getState().examState;
+  const { exam } = store.getState().specialExams;
   const { dispatch } = store;
   if (!exam) {
     return Promise.resolve();
--- a/src/instructions/onboarding_exam/EntranceOnboardingExamInstructions.jsx
+++ b/src/instructions/onboarding_exam/EntranceOnboardingExamInstructions.jsx
@@ -1,11 +1,15 @@
-import React, { useContext } from 'react';
+import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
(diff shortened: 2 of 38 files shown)


src/api.js
```diff
@@ -1,20 +1,20 @@
 import { examRequiresAccessToken, store } from './data';
 
 export function isExam() {
-  const { exam } = store.getState().examState;
+  const { exam } = store.getState().specialExams;
```
varshamenon4, line 4: [comment] I like the renaming to specialExams!

src/instructions/onboarding_exam/EntranceOnboardingExamInstructions.jsx
```diff
@@ -89,7 +93,7 @@ const EntranceOnboardingExamInstructions = () => {
         <Button
           data-testid="start-exam-button"
           variant="primary"
-          onClick={createProctoredExamAttempt}
+          onClick={() => dispatch(createProctoredExamAttempt())}
```
varshamenon4, line 96: [question] For the codecov, I think you'd mentioned coming back to this as a larger effort to clean up / update testing, right? So for now, not worrying about codecov?

varshamenon4 (approved):
Woah this is awesome! Lots of refactoring - looks great!

src/instructions/onboarding_exam/EntranceOnboardingExamInstructions.jsx
MichaelRoytman: Yeah, I think so. The tests in here are kind of difficult to work with. I think we want to get all of our code changes into the feature branch, and then we'll do a mini bug bash and add any missing tests. Does that sound good to you?

PR #132, opened 2024-02-01
feat: refactor non instruction components

alangsto:
## [COSMO-172](https://2u-internal.atlassian.net/browse/COSMO-172)

This PR removes the use of the `ExamStateProvider` component and `withExamStore` function, replacing any calls to context or state with a selector. 

The `examStore` has also been renamed to `specialExams` in preparation for our eventual merger of the exams store into the learning MFE.


alangsto:
The public API test suite is now failing, but the work to fix those tests + API will be part of COSMO-175

src/core/OuterExamTimer.jsx
```diff
@@ -1,23 +1,32 @@
...
   const {
-    activeAttempt, showTimer, stopExam, submitExam,
-    expireExam, pollAttempt, apiErrorMsg, pingAttempt,
-    getLatestAttemptData,
-  } = state;
+    activeAttempt, apiErrorMsg,
+  } = useSelector(state => state.specialExams);
+  const dispatch = useDispatch();
+
+  const showTimer = !!(activeAttempt && IS_STARTED_STATUS(activeAttempt.attempt_status));
```
varshamenon4, line 25: TIL what !! is - cooool! Love "bang, bang you're boolean" here: https://stackoverflow.com/questions/784929/what-is-the-not-not-operator-in-javascript
Nice!

src/core/OuterExamTimer.jsx
```diff
@@ -31,11 +40,11 @@ const ExamTimer = ({ courseId }) => {
...
-          stopExamAttempt={stopExam}
-          submitExam={submitExam}
-          expireExamAttempt={expireExam}
-          pollExamAttempt={pollAttempt}
-          pingAttempt={pingAttempt}
+          stopExamAttempt={() => dispatch(stopExam())}
+          submitExam={() => dispatch(submitExam())}
+          expireExamAttempt={() => dispatch(expireExam())}
+          pollExamAttempt={(url) => dispatch(pollAttempt(url))}
+          pingAttempt={(timeout, url) => dispatch(pingAttempt(timeout, url))}
```
varshamenon4, line 47: Nice!

src/exam/Exam.jsx
```diff
@@ -70,6 +77,7 @@ const Exam = ({
     // this makes sure useEffect gets called only one time after the exam has been fetched
     // we can't leave this empty since initially exam is just an empty object, so
     // API calls above would not get triggered
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
varshamenon4, line 80: [question] I noticed this in a couple of places. Could you clarify the need for this / if this will eventually be unnecessary?

varshamenon4 (commented):
This is awesome! Definitely really complicated stuff. Just a couple of comments and questions.

src/context.jsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
-
-const ExamStateContext = React.createContext({});
-export default ExamStateContext;
+// import React from 'react';
```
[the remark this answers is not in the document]
alangsto: I removed them now!

src/exam/Exam.jsx
alangsto: This lint rule is meant to try and prevent developers from forgetting dependencies in the use effect. Because I didn't change anything within the use effect, I'm planning to leave this as is for now!

src/core/OuterExamTimer.jsx
```diff
@@ -2,16 +2,15 @@ import React, { useEffect, useContext } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import { AppContext } from '@edx/frontend-platform/react';
-import ExamStateContext from '../context';
-import { getLatestAttemptData } from '../data';
 import { ExamTimerBlock } from '../timer';
 import ExamAPIError from '../exam/ExamAPIError';
-import ExamStateProvider from './ExamStateProvider';
+import { getLatestAttemptData } from '../data/thunks';
```
MichaelRoytman, line 7: Nit. It seems like most components import from the top level `data` module (things exported from `data/index.js`). Should we follow that pattern here and in the other files?

src/data/store.js
```diff
@@ -1,8 +1,8 @@
-import { configureStore } from '@reduxjs/toolkit';
-import examReducer from './slice';
-
-export default configureStore({
-  reducer: {
-    specialExams: examReducer,
-  },
-});
+// import { configureStore } from '@reduxjs/toolkit';
```
MichaelRoytman, line 1: I think we're okay to remove this, since we're still on a feature branch, right? Or were you thinking of doing that in a follow up pull request?

src/exam/Exam.jsx
```diff
@@ -23,10 +23,12 @@ import messages from './messages';
...
 }) => {
-  const state = useContext(ExamStateContext);
   const {
-    isLoading, showTimer, exam, apiErrorMsg, getProctoringSettings,
-  } = state;
+    isLoading, activeAttempt, exam, apiErrorMsg,
+  } = useSelector(state => state.specialExams);
+  const dispatch = useDispatch();
+
+  const showTimer = !!(activeAttempt && IS_STARTED_STATUS(activeAttempt.attempt_status));
```
MichaelRoytman, line 31: Nit. If you wanted to reduce some of the duplication in the definition of `showTimer`, you could put it in a custom hook.

src/exam/Exam.jsx
MichaelRoytman: Should it be removed?

I'm wondering if it's complaining because you're using `dispatch` in the `useEffect` hook, so `dispatch` has to be added to the dependency array.

MichaelRoytman (approved):
I left a few nits, but this looks great 🎉!

src/exam/Exam.jsx
alangsto: Oh, fair point! I will try adding that in and see what happens.

src/core/OuterExamTimer.jsx
alangsto: True, will update!

src/data/store.js
alangsto: Let me try removing that as part of this one

alangsto: Removing the store requires quite a bit of refactoring, because most of the tests directly import the store. I'll work through that now though!

MichaelRoytman: Oof! I didn't realize that. If you want to leave that to another ticket, this pull request looks good to merge to me, but I'm happy to re-review if you want to refactor the tests.

alangsto: It wasn't as bad as I expected because I could just use the test store as a replacement! Thankful that it existed. Should be set for another review now :)

src/data/store.js
```diff
@@ -1,8 +0,0 @@
-import { configureStore } from '@reduxjs/toolkit';
```
MichaelRoytman, line 1: 🎉

MichaelRoytman (approved):

PR #133, opened 2024-02-01
Refactor TimerServiceProvider to no longer use withExamStore higher-order-component.

MichaelRoytman:
### Description:

**Jira**: [COSMO-174](https://2u-internal.atlassian.net/browse/COSMO-174)

* refactor: replace use of withExamStore higher-order component in TimerServiceProvider
  * This commit replaces the use of the withExamStore higher-order component with the useDispatch and useSelector hooks in TimerServiceProvider.
  * This commit also refactors components that use the TimerServiceProvider so that they no longer need to pass state and action creators via props. The TimerServiceProvider now gets whatever state it needs directly from the Redux store with a useSelector hook and imports and dispatches thunks directly by importing them from the data directory.
  * The original pattern was to use the withExamStore higher-order component to provide context to the TimerServiceProvider and its children. This context contained provided the Redux store state and action creators as props by using the connect API. This posed a problem for our need to merge the frontend-app-learning and frontend-lib-special-exams stores, because the special exams store is initialized in this repository and used by the higher-order component. In order to eventually be able to remove the creation of the store in this repository, we have to remove references to the store by interfacing with the Redux more directly by using the useDispatch and useSelector hooks.

### DIFF
--- a/src/core/OuterExamTimer.jsx
+++ b/src/core/OuterExamTimer.jsx
@@ -1,22 +1,24 @@
 import React, { useEffect, useContext } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import { AppContext } from '@edx/frontend-platform/react';
 import ExamStateContext from '../context';
+import { getLatestAttemptData } from '../data';
 import { ExamTimerBlock } from '../timer';
 import ExamAPIError from '../exam/ExamAPIError';
 import ExamStateProvider from './ExamStateProvider';
 
 const ExamTimer = ({ courseId }) => {
-  const state = useContext(ExamStateContext);
+  const examState = useContext(ExamStateContext);
   const { authenticatedUser } = useContext(AppContext);
-  const {
-    activeAttempt, showTimer, stopExam, submitExam,
-    expireExam, pollAttempt, apiErrorMsg, pingAttempt,
-    getLatestAttemptData,
-  } = state;
+  const { showTimer } = examState;
+
+  const { apiErrorMsg } = useSelector(state => state.specialExams);
+
+  const dispatch = useDispatch();
 
   useEffect(() => {
-    getLatestAttemptData(courseId);
+    dispatch(getLatestAttemptData(courseId));
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [courseId]);
 
@@ -29,14 +31,7 @@ const ExamTimer = ({ courseId }) => {
   return (
     <div className="d-flex flex-column justify-content-center">
       {showTimer && (
-        <ExamTimerBlock
-          attempt={activeAttempt}
-          stopExamAttempt={stopExam}
-          submitExam={submitExam}
-          expireExamAttempt={expireExam}
-          pollExamAttempt={pollAttempt}
-          pingAttempt={pingAttempt}
-        />
+        <ExamTimerBlock />
       )}
       {apiErrorMsg && <ExamAPIError />}
     </div>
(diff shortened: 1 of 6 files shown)


src/core/OuterExamTimer.jsx
```diff
@@ -1,22 +1,24 @@
...
 const ExamTimer = ({ courseId }) => {
-  const state = useContext(ExamStateContext);
+  const examState = useContext(ExamStateContext);
   const { authenticatedUser } = useContext(AppContext);
-  const {
-    activeAttempt, showTimer, stopExam, submitExam,
-    expireExam, pollAttempt, apiErrorMsg, pingAttempt,
-    getLatestAttemptData,
-  } = state;
+  const { showTimer } = examState;
```
alangsto, line 14: showTimer is only available if ExamStateProvider is being used, which we want to move away from. I did some of this clean up in https://github.com/openedx/frontend-lib-special-exams/pull/132, but I like how you removed all the props being passed down to the child Timer components. 

We might need to do a mix of both your changes and my changes.

src/core/OuterExamTimer.jsx
```diff
@@ -1,22 +1,24 @@
...
 import { AppContext } from '@edx/frontend-platform/react';
 import ExamStateContext from '../context';
+import { getLatestAttemptData } from '../data';
 import { ExamTimerBlock } from '../timer';
 import ExamAPIError from '../exam/ExamAPIError';
 import ExamStateProvider from './ExamStateProvider';
 
 const ExamTimer = ({ courseId }) => {
-  const state = useContext(ExamStateContext);
+  const examState = useContext(ExamStateContext);
```
varshamenon4, line 12: Nice! I like the specificity of using examState here.

src/exam/Exam.jsx
```diff
@@ -25,9 +25,7 @@ const Exam = ({
 }) => {
   const state = useContext(ExamStateContext);
   const {
-    isLoading, activeAttempt, showTimer, stopExam, exam,
-    expireExam, pollAttempt, apiErrorMsg, pingAttempt,
-    getProctoringSettings, submitExam,
+    isLoading, showTimer, exam, apiErrorMsg, getProctoringSettings,
```
alangsto, line 28: Same issue with showTimer here

src/core/OuterExamTimer.jsx
```diff
@@ -1,22 +1,24 @@
...
   const { authenticatedUser } = useContext(AppContext);
-  const {
```
varshamenon4, line 12: [question] Do we just no longer need all of these consts from state?

src/exam/Exam.jsx
alangsto: Although maybe it's not an issue? I just am not sure why we need ExamStateContext anymore

varshamenon4 (approved):
woohoo, lgtm!

src/exam/Exam.jsx
MichaelRoytman: Just reiterating our conversation from standup for posterity, but I was just trying not to interfere with the files you were working on. I agree that we can move `showTimer` into the components that need it and get rid of `ExamStateProvider`!

src/core/OuterExamTimer.jsx
MichaelRoytman: If you take a look at where these were being used in this component, they were just being passed as props to `ExamTimerBlock` and weren't being used anywhere else (except `showTimer`).

I've refactored `ExamTimerBlock` to pull these values either out of Redux (if they're state) or import them out of the `data` module (if they're thunks), which means `ExamTimerBlock` no longer needs to get these values as props. `OuterExamTimer` was getting these as props from _its_ parent and doing some prop drilling to send them to `ExamTimerBlock`, but since `ExamTimerBlock` can now get what it needs itself, it no longer needs `OuterExamTimer` to send it those props.

PR #135, opened 2024-02-07
fix: patch bugs found during bash

alangsto:
There were a few things we missed in the refactoring that we caught during our first bug bash. These fixes ensure that an exam will be autosubmitted if the time runs out, and that an exam can be started.

### DIFF
--- a/src/timer/ExamTimerBlock.jsx
+++ b/src/timer/ExamTimerBlock.jsx
@@ -6,7 +6,7 @@ import CountDownTimer from './CountDownTimer';
 import { ExamStatus, IS_STARTED_STATUS } from '../constants';
 import TimerProvider from './TimerProvider';
 import {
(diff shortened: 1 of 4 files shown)


src/timer/ExamTimerBlock.jsx
```diff
@@ -6,7 +6,7 @@ import CountDownTimer from './CountDownTimer';
 import { ExamStatus, IS_STARTED_STATUS } from '../constants';
 import TimerProvider from './TimerProvider';
 import {
-  Emitter, expireExamAttempt, stopExam, submitExam,
+  Emitter, expireExam, stopExam, submitExam,
```
MichaelRoytman, line 9: I don't think anything else in the library uses the `expireExamAttempt` function, so I think we should remove the export from `data/index.js` as well.

MichaelRoytman (approved):
Thank you! 🎉

PR #136, opened 2024-02-08
2u/cosmonauts/refactor store

alangsto:
## [COSMO-85](https://2u-internal.atlassian.net/browse/COSMO-85)

In order to gate chat visibility in the learning MFE by whether or not a learner is viewing an exam or has an active exam attempt, the exam state from `frontend-lib-special-exams` needed to be merged with the learning MFE store. 

This PR is the culmination of several smaller commits that were all previously reviewed. Each commit contains code changes to refactor this library from using a HOC to pass state to child components to instead use selectors to read directly from the store. 

@MichaelRoytman and I both tested this locally, and were able to successfully complete a timed exam.


src/api.test.jsx
```diff
@@ -1,60 +1,80 @@
 import { Factory } from 'rosie';
 
-import { isExam, getExamAccess, fetchExamAccess } from './api';
-import { store } from './data';
+import { useExamAccessToken, useFetchExamAccessToken, useIsExam } from './api';
+import { initializeTestStore, render } from './setupTest';
+
+/**
```
varshamenon4, line 6: Nice docs!

src/instructions/proctored_exam/ReadyToStartProctoredExamInstructions.jsx
```diff
@@ -1,31 +1,31 @@
...
   const examDuration = attempt.total_time ? attempt.total_time : exam.total_time;
   const platformName = getConfig().SITE_NAME;
   const rulesUrl = getConfig().PROCTORED_EXAM_RULES_URL;
   const [beginExamClicked, setBeginExamClicked] = useState(false);
 
   useEffect(() => {
-    getExamReviewPolicy();
+    dispatch(getExamReviewPolicy());
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
```
MichaelRoytman, line 24: I found it odd we didn't get a lint error here for missing `dispatch`. I did a little reading, and it seems like the reference to `dispatch` is [very unlikely to change](https://stackoverflow.com/a/67012349). Maybe some linters have caught up to this and no longer lint on this? Where did you see that linting error originally?

I ask because I'd prefer we're consistent about whether we have dispatch in the dependency array.

varshamenon4 (approved):
Woah, this is incredible! Nice! lgtm!

MichaelRoytman (approved):

src/instructions/proctored_exam/ReadyToStartProctoredExamInstructions.jsx
alangsto: I believe the reason we're not getting a lint error is because of the eslint-disable comment on the next line. I agree that we should be consistent, so I can add `dispatch` to the dependency array.

PR #138, opened 2024-02-12
feat: pass legacy status of exam to failure thunk

alangsto:
## [COSMO-182](https://2u-internal.atlassian.net/browse/COSMO-182)

All thunks which result in a call to update an attempt status should account for whether or not the exam attempt is for a non-LTI provider. If the legacy status is not passed to the thunk, in this case `endExamWithFailure`, the frontend will only call `edx-exams`, which does not support routing PUT requests to update an attempt status back to `edx-proctoring`.


varshamenon4 (approved):
lgtm!

PR #139, opened 2024-02-22
fix: pass sequenceId to api call

ilee2u:
Fix for https://2u-internal.atlassian.net/browse/COSMO-125

We seemed to forget to pass the sequenceId to the backend API call in thunks.js. I just added that back in, hopefully this should work on stage now.

### DIFF
--- a/src/data/__snapshots__/redux.test.jsx.snap
+++ b/src/data/__snapshots__/redux.test.jsx.snap
@@ -203,6 +203,7 @@ Object {
   "software_download_url": "",
   "taking_as_proctored": false,
   "time_remaining_seconds": 1739.9,
+  "timer_ends": "2024-03-11T17:31:55.297Z",
   "total_time": "30 minutes",
 }
 `;
--- a/src/data/__snapshots__/redux.test.jsx.snap
+++ b/src/data/__snapshots__/redux.test.jsx.snap
@@ -188,7 +188,7 @@ Object {
 }
 `;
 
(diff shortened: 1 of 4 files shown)


Zacharis278:
this does just seem like a missed piece of code. One question tho, does this work correctly if the user is outside of the exam section? Like on the course outline or viewing a different piece of course content?

alangsto (approved):

ilee2u:
> this does just seem like a missed piece of code. One question tho, does this work correctly if the user is outside of the exam section? Like on the course outline or viewing a different piece of course content?

I tried messing around with this on stage, going between the course view and the exam sequence view, which actually didn't cause any problems. In fact, I only saw the error in the exam sequence view, and interestingly, going between both views caused the problem not to happen.

Also, I determined that this was probably happening because of pollExamAttempt due to it always happening when the timer hit the :00 minute mark, meaning that this was going off every 60 seconds when pollAttempt is called.

Zacharis278:
@rijuma this might have been the cause of your bug as well.

@ilee2u have we tested those cases with this fix added? Wanna make sure the state we're pulling has the correct content id for the active attempt if this happens to get called when not in the right sequence.

schenedx (approved):
👍

Zacharis278:
@ilee2u did we ever confirm this works when the user is outside of the particular exam section? I just tried a timed exam on stage and I get an error when viewing the outline with the timer running.

PR #140, opened 2024-02-23
Fixed exam remaining time display offset

rijuma:
## Description
Ticket: [COSMO-180 🔒](https://2u-internal.atlassian.net/browse/COSMO-180)

While taking an exam, the remaining time displayed in the exam timer is inconsistent. It gets an increasing offset when navigating around in a course in a way that requires a browser refresh. It appears that this does not occur when the browser does not refresh (e.g. navigating between units in the same subsection via the navigation bar). It’s usually up to a 40 second difference in either direction.

### Cause
The cause for this offset was caused by the fact that we weren't using a fixed time reference for the remaining time. The endpoint returns the remaining time in seconds which was saved in the redux store, and then when the TimerProvider was rendered it was calculating the remaining time based on the current time. The first render would be accurate but after that each re-render of the component would take the same value to start the timer. This caused an offset which eventually would be corrected by a poll that happens every minute, which mitigates but does not solve the problem.

### Proposed solution
In order to fix this, my approach is to calculate a `timerEnds` date at the moment we store this information in Redux, and then use this timer to calculate the seconds to show to the user. Since the date is fixed in time, no offset is created no matter the reloads.

## Extra changes
There are some minor issues that I detected wile working on this. For example, sometimes the frontend wanted to poll periodically to an url that was returned by the endpoint `exam_started_poll_url`, which was indicated by a comment in the code `poll url may be null if this is an LTI exam`. In my experience, these cases were showing an obscure error to the user.

Screenshot 2024-02-23 at 10 26 17

I've used this opportunity to reshape the error in a more descriptive way:

Screenshot 2024-02-23 at 10 28 49

We are now avoiding the poll when the url is not provided to avoid this error overall, but I thought it was a nice thing to keep just in case.

## Update (2024-02-27)
- When updating the tests I've added `act()` to wrap the `await expect()` to wait for update changes, since there are state changes involved and jest expects us to use `act()` when this happens.
- Added some tweaks that might raise an eyebrow, especially on the timeless `useTimeout()` to wait for the children render to subscribe to the emitter. Please check it when reviewing.

## Update (2024-02-29)
- Added 100% coverage to TimeProvider. It didn't had any coverage before.
- Made some extra tweaks to the code while testing. I feel pretty confident that it works as it should.
- Note: There's a weird choice on how the ping was working, if the `ping_interval` is set, it runs first at half of the time, and then each full interval. I left it as it is, but I wanted to bring attention to it.
- Added a `test:watch` to `package.json`'s scripts that's handy when working on new tests. The coverage output gets too much in the way.


src/timer/TimerProvider.jsx
```diff
@@ -52,15 +49,19 @@ const TimerProvider = ({
     },
   ).join(':');
 
-  const pollExam = () => {
+  const pollExam = useCallback(() => {
```
rijuma, line 52: This function was created each render, which happens each second. the `useCallback()` is to prevent this for efficiency.

src/timer/TimerProvider.jsx
```diff
@@ -52,15 +49,19 @@ const TimerProvider = ({
...
     // poll url may be null if this is an LTI exam
-    dispatch(pollAttempt(attempt.exam_started_poll_url));
-  };
+    if (attempt?.exam_started_poll_url) {
+      dispatch(pollAttempt(attempt.exam_started_poll_url));
+    }
+  }, [attempt.exam_started_poll_url, dispatch]);
+  const processTimeLeft = useCallback(() => (secondsLeft) => {
```
rijuma, line 58: Same as before, the function was created each render, which happens each second. the `useCallback()` is to prevent this for efficiency.

src/timer/TimerProvider.jsx
```diff
@@ -52,15 +49,19 @@ const TimerProvider = ({
...
+    const criticalLowTime = timeLimitMins * 60 * TIME_LIMIT_CRITICAL_PCT;
+    const lowTime = timeLimitMins * 60 * TIME_LIMIT_LOW_PCT;
```
rijuma, line 60: These constants were renamed and relocated since they are not actual app constants, but just calculations based on context. The uppercase naming was misleading.

src/timer/TimerProvider.jsx
```diff
@@ -69,22 +70,35 @@ const TimerProvider = ({
     if (secondsLeft <= 0) {
       Emitter.emit(TIMER_REACHED_NULL);
     }
+
     if (!limitReached && secondsLeft < LIMIT) {
-      clearInterval(timer);
```
rijuma, line 73: I've moved this to be handled in the interval block, since I find it easier to understand.

src/timer/TimerProvider.jsx
```diff
@@ -28,20 +29,16 @@ const TimerProvider = ({
...
     desktop_application_js_url: workerUrl,
     ping_interval: pingInterval,
-    time_remaining_seconds: timeRemaining,
+    time_remaining_seconds: initSecondsLeft,
+    timer_ends: timerEnds,
   } = attempt;
-  const LIMIT = GRACE_PERIOD_SECS ? 0 - GRACE_PERIOD_SECS : 0;
-  const CRITICAL_LOW_TIME = timeLimitMins * 60 * TIME_LIMIT_CRITICAL_PCT;
-  const LOW_TIME = timeLimitMins * 60 * TIME_LIMIT_LOW_PCT;
-  let liveInterval = null;
```
rijuma, line 42: The control variable was moved to the useEffect block, since the component refresh would overwrite this value rendering it useless.

src/helpers.js
```diff
@@ -36,3 +36,22 @@ export const generateHumanizedTime = (timeRemainingSeconds) => {
   }
   return remainingTime;
 };
+
+// The only information we get on the remaining time on the active exam attempt
+// from the endpoint is the remaining seconds. We need to have a fixed time reference
+// on the time limit to be able to calculate the remaining time accurately.
+export const aggregateActiveAttemptData = (activeAttempt) => {
```
schenedx, line 43: This function name `aggregateActiveAttemptData` is confusing. Why use the word `aggregate`? 
This function essentially appends the `timer_ends` attribute to the original json. Why not just call the function `appendTimerEnd`?

src/timer/TimerProvider.jsx
```diff
@@ -93,14 +106,27 @@ const TimerProvider = ({
       if (workerUrl && timerTick % pingInterval === pingInterval / 2) {
         dispatch(pingAttempt(pingInterval, workerUrl));
       }
+
+      const keepRunning = processTimeLeft(secondsLeft);
```
schenedx, line 110: Can you add some comments here what case are you handling in checking this condition? Is it when timer finished running for the whole exam?

Zacharis278:
Great, the approach looks pretty good. I'd take a look at @ilee2u's PR about the error cropping up. https://github.com/openedx/frontend-lib-special-exams/pull/139. I think both of these problems are slightly related since it wasn't always hitting the backend to sync up like it should have

src/helpers.js
rijuma: By ´aggregate´ I meant that I'm adding more data to existing data. I think I've misused the term, maybe something like ´compute´ or ´process´ would have been better. I tend to name functions in a generic way, since this can be used to add more data in the future. I'll rename it as ´appendTimerEnd´ instead as suggested.

src/timer/TimerProvider.jsx
```diff
@@ -52,15 +48,19 @@ const TimerProvider = ({
     },
   ).join(':');
 
-  const pollExam = () => {
+  const pollExam = useCallback(() => {
     // poll url may be null if this is an LTI exam
-    dispatch(pollAttempt(attempt.exam_started_poll_url));
-  };
+    if (attempt?.exam_started_poll_url) {
+      dispatch(pollAttempt(attempt.exam_started_poll_url));
```
Zacharis278, line 54: we do need this to keep polling even if the url param is 'missing'. The explanation for this one is a bit lengthy but basically this UI code works with either edx-exam or edx-proctoring as the backedn. The API for the former is slightly different in that we don't use this polling callback URL and instead have a static route that gets called. However there is a bug here getting fixed by https://github.com/openedx/frontend-lib-special-exams/pull/139

src/timer/TimerProvider.jsx
rijuma: I've reamed the var as `keepTimerRunning` for the purpose of it to be clearer.

package.json
```diff
@@ -31,7 +31,8 @@
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "lint:fix": "fedx-scripts eslint --fix --ext .js --ext .jsx .",
     "snapshot": "fedx-scripts jest --updateSnapshot",
-    "test": "fedx-scripts jest --coverage --passWithNoTests"
+    "test": "fedx-scripts jest --coverage --passWithNoTests",
+    "test:watch": "fedx-scripts jest --passWithNoTests --watch"
```
rijuma, line 35: I've found this to be very practical when working on tests. The coverage table gets too much in the way.

src/data/thunks.js
```diff
@@ -272,6 +272,9 @@ export function pollAttempt(url) {
 
     try {
       const data = await pollExamAttempt(url);
+      if (!data) {
+        throw new Error('Poll Exam failed to fetch.');
```
rijuma, line 276: When for some reason the poll exam attempt fetch returns undefined, a really annoying error shows as:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'time_remaining_seconds').
```
Which is very unpractical. I'd rather have some better context if this happens.

package.json
```diff
@@ -89,4 +90,4 @@
     "rosie": "2.0.1",
     "semantic-release": "^20.1.3"
   }
-}
+}
```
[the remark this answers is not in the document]
rijuma: Fixed. I had to update my IDE to let me add it on json files for some reason.. Thanks for noticing it.

src/data/redux.test.jsx
```diff
@@ -53,6 +53,18 @@ describe('Data layer integration tests', () => {
     await executeThunk(thunks.getExamAttemptsData(courseId, contentId), store.dispatch);
   };
 
+  const expectSpecialExamAttemptToMatchSnapshot = (data) => expect(data).toMatchSnapshot({
```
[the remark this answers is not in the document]
rijuma: I've added a short comment:
```javascript
// This is a shorthand to check snapshots with an assymetric matcher every time.
```
Let me know if this works for you,

PR #142, opened 2024-03-14
fix: workaround for pollAttempt fail in courseview

ilee2u:
Re-fix for: https://2u-internal.atlassian.net/browse/COSMO-125

- Combined fetchActiveAttempt &  fetchLatestExamAttempt into one function with an optional "sequenceId" variable. At the very least, this works without causing errors in non-exam-sequence views. Since the sequenceId is undefined in that case, due to the sequenceId needing to be extracted from the sequence page itself, we can at least still get the latest exam attempt data.

- To my best knowledge, we originally called for the attempt of an exam with a given sequenceId in order to get the latest attempt for an exam regardless of whether or not its active, such that the timer can be seen on non exam-sequence pages. However, we don't truly need the sequence_id for this.

- It's also possible that this was also designed this way so that we could make sure a learner hadn't started two exams at once, and so their "latest exam attempt" would pretain to the exam they started most recently. We already have safeguards for this in edx_exams (See: https://github.com/edx/edx-exams/blob/main/edx_exams/apps/core/api.py#L201).

EDIT: I have not updated my tests yet because I want to make sure this approach is good first

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -18,6 +18,17 @@ async function fetchActiveAttempt(sequenceId = null) {
   return activeAttemptResponse.data;
 }
 
+async function fetchAttemptForExamSequnceId() {
+    // fetch 'active' (timer is running) attempt if it exists
+    const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
+    const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
+    // the calls the same endpoint as fetchActiveAttempt but it behaves slightly different
+    // with an exam's section specified. The attempt for that requested exam is always returned
+    // even if it is not 'active' (timer is not running)
+    activeAttemptUrl.searchParams.append('content_id', sequenceId);
+    return activeAttemptResponse.data;
+}
+
 export async function fetchExamAttemptsData(courseId, sequenceId) {
   let data;
   if (!getConfig().EXAMS_BASE_URL) {
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -5,28 +5,21 @@ import { generateHumanizedTime } from '../helpers';
 
 const BASE_API_URL = '/api/edx_proctoring/v1/proctored_exam/attempt';
 
-async function fetchActiveAttempt(sequenceId = null) {
+async function fetchActiveAttempt() {
   // fetch 'active' (timer is running) attempt if it exists
   const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
   const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
-  if (sequenceId) {
-    // the calls the same endpoint as fetchActiveAttempt but it behaves slightly different
-    // with an exam's section specified. The attempt for that requested exam is always returned
-    // even if it is not 'active' (timer is not running)
-    activeAttemptUrl.searchParams.append('content_id', sequenceId);
-  }
   return activeAttemptResponse.data;
 }
 
-async function fetchAttemptForExamSequnceId() {
-    // fetch 'active' (timer is running) attempt if it exists
-    const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
-    const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
-    // the calls the same endpoint as fetchActiveAttempt but it behaves slightly different
-    // with an exam's section specified. The attempt for that requested exam is always returned
-    // even if it is not 'active' (timer is not running)
-    activeAttemptUrl.searchParams.append('content_id', sequenceId);
-    return activeAttemptResponse.data;
+async function fetchAttemptForExamSequnceId(sequenceId) {
+  const attemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
+  // the calls the same endpoint as fetchActiveAttempt but it behaves slightly different
+  // with an exam's section specified. The attempt for that requested exam is always returned
+  // even if it is not 'active' (timer is not running)
+  attemptUrl.searchParams.append('content_id', sequenceId);
+  const attemptResponse = await getAuthenticatedHttpClient().get(attemptUrl.href);
+  return attemptResponse.data;
 }
 
 export async function fetchExamAttemptsData(courseId, sequenceId) {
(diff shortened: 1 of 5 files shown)


src/data/thunks.js
```diff
@@ -271,8 +271,7 @@ export function pollAttempt(url) {
     }
 
     try {
-      // TODO: make sure sequenceId pulled here is correct both in-exam-sequence and in outline
-      // test w/ timed exam
+      // Why is this undefined in the course view? Let's see how it's called there...
```
alangsto, line 274: Did you mean to include this comment?

src/data/api.js
```diff
@@ -70,25 +65,27 @@ export async function fetchLatestAttempt(courseId) {
...
       `${getConfig().LMS_BASE_URL}${pollUrl}`,
     );
     const urlResponse = await getAuthenticatedHttpClient().get(edxProctoringURL.href);
     data = urlResponse.data;
-  } else if (sequenceId && getConfig().EXAMS_BASE_URL) {
-    data = await fetchLatestExamAttempt(sequenceId);
 
+  // sites configured with edx-exams expect sequenceId if pollUrl is not set
+  // and the learner is viewing the exam sequence
+  // (the exam sequence is not consumed outside the exam sequence, e.g. in the course view)
```
alangsto, line 79: I think this comment could be a bit clearer, specifically the piece in parenthesis.

src/data/api.js
```diff
@@ -1,28 +1,23 @@
...
 
-async function fetchActiveAttempt() {
+async function fetchActiveAttempt(sequenceId = null) {
   // fetch 'active' (timer is running) attempt if it exists
   const activeAttemptUrl = new URL(`${getConfig().EXAMS_BASE_URL}/api/v1/exams/attempt/latest`);
   const activeAttemptResponse = await getAuthenticatedHttpClient().get(activeAttemptUrl.href);
+  if (sequenceId) {
+    // the calls the same endpoint as fetchActiveAttempt but it behaves slightly different
+    // with an exam's section specified. The attempt for that requested exam is always returned
+    // even if it is not 'active' (timer is not running)
```
alangsto, line 15: I think this comment should be updated as well, because there is only one `fetchActiveAttempt` function, so the wording here doesn't make that much sense anymore.

src/data/api.js
```diff
@@ -1,28 +1,23 @@
 import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
-import { logError } from '@edx/frontend-platform/logging';
 import { ExamAction } from '../constants';
 import { generateHumanizedTime } from '../helpers';
 
 const BASE_API_URL = '/api/edx_proctoring/v1/proctored_exam/attempt';
 
-async function fetchActiveAttempt() {
+async function fetchActiveAttempt(sequenceId = null) {
```
alangsto, line 8: Is there test coverage for the updated function now?

src/data/api.js
```diff
@@ -70,25 +65,27 @@ export async function fetchLatestAttempt(courseId) {
...
+  // and the learner is viewing the exam sequence
+  // (the exam sequence is not consumed outside the exam sequence, e.g. in the course view)
+  } else {
+    data = await fetchActiveAttempt(sequenceId);
     // Update dictionaries returned by edx-exams to have correct status key for legacy compatibility
     if (data.attempt_status) {
       data.status = data.attempt_status;
       delete data.attempt_status;
     }
-  } else {
```
Zacharis278, line 87: It would be good to keep similar parameter assertion. It might not be obvious to troubleshoot or even notice if a change in a component causes a problem here.

Zacharis278 (commented):
I think there might be a root issue with the sequence id we're pulling here: https://github.com/openedx/frontend-lib-special-exams/blob/main/src/data/thunks.js#L277

This change handles the sequence id being none on the outline but what if the learner navigates elsewhere in the course?

Zacharis278:
@ilee2u I had a thought on this that might help if it turns out passing the correct sequence id is not simple.

The cause of this bug is this PR here https://github.com/openedx/frontend-lib-special-exams/pull/113 that assumes pollExamAttempt is always called with a sequence id (either in the pollUrl or directly). I think the simple fix here is adding a 3rd case to this function where we call fetchActiveAttempt instead of fetchLatestAttempt if no sequence (or poll url) is provided.

ilee2u:
I discovered that [this line in TimerProvider](https://github.com/openedx/frontend-lib-special-exams/blob/main/src/timer/TimerProvider.jsx#L146) in frontend-lib-special-exams causes the OuterExamTimer to ping the exams backend non-stop, as its value changes every time the exam is polled, which causes pollExam to be called as its in the dependency array of a useEffect(), causing a vicious cycle.

It seems like just removing this one line causes things to work again, though removing it triggers lint since it's "missing" from the dependency array, so I added a lint ignore

src/data/redux.test.jsx
```diff
@@ -990,16 +991,21 @@ describe('Data layer integration tests', () => {
...
+        });
+        await api.pollExamAttempt(null);
+        expect(axiosMock.history.get[0].url).toEqual(expectedUrl);
       });
     });
   });
 
   describe('Test pingAttempt', () => {
     it('Should send attempt to error state on ping failure', async () => {
+      // FIXME:
```
schenedx, line 1008: Why are these comments in the PR?

ilee2u: Ah I left those in by accident, let me remove those real quick

src/timer/TimerProvider.jsx
```diff
@@ -142,8 +142,7 @@ const TimerProvider = ({
       clearInterval(timerRef.current);
       timerRef.current = null;
     };
-  }, [
-    timerEnds,
+  }, [ // eslint-disable-line react-hooks/exhaustive-deps
```
MichaelRoytman, line 145: My understanding of `timerEnds` is that it _should_ have a consistent value across calls to `pollAttempt`. Its purpose is to be a fixed time reference. What are the values of `timerEnds` that you're seeing locally that are not equal? I'd like for you or Marcos to double-check me on my understanding, though!

Unfortunately, written as is, this `useEffect` hook needs to have `timerEnds` in the dependency array, because it's used within the hook; see [Specifying Reactive dependencies](https://react.dev/reference/react/useEffect#specifying-reactive-dependencies). If we were to move `deadline` outside of the hook, we could remove `timerEnds` from the dependency array. I _think_ that should be fine, assuming `timerEnds` represents a static time.

rijuma: Yes, I've mentioned this on our [slack thread](https://twou.slack.com/archives/C049EF6VCTB/p1710959512096379). This was an oversight from me when I worked on the `TimerProvider`. Removing the dependency is not the best approach.

I think the best would be to move this dependency to a reference so it does not need to be added as a dependency and if it gets updated it will not re-run the `useEffect`. We should just make sure that we are using the reference instead of the primitive to avoid any issues.

rijuma: To reply the question above:
>  What are the values of timerEnds that you're seeing locally that are not equal? I'd like for you or Marcos to double-check me on my understanding, though!

The problem here is caused by `timerEnds` being recalculated in each poll. Since the endpoint returns only the remaining seconds, this value can mean different things depending the time is evaluated. In this case, since `timerEnds` gets calculated when is saved at the store, this value probably gets a slightly minor offset every time (depending on network and execution delay), so the value is slightly different and the hook reacts to it.

This was the original problem for ticket [COSMO-180](https://2u-internal.atlassian.net/browse/COSMO-180) since every render it would take the remaining seconds again from the store based on the poll time and not the current time.

Is there a possibility to update the endpoint to return a datetime instead of the remaining seconds? Although this would mean that we need to deal with time offsets...

This problem could use some brainstorming.

ilee2u: Marcos paired with me and we were able to fix this :)

src/timer/TimerProvider.jsx
```diff
@@ -95,15 +95,20 @@ const TimerProvider = ({
     timeLimitMins,
   ]);
 
+  // Set deadline as a reference to timerEnds that updates when it changes
+  const deadline = useRef(new Date(timerEnds));
+  useEffect(() => {
+    deadline.current = new Date(timerEnds);
+  }, [timerEnds]);
```
rijuma, line 102: Thank you for fixing this!

rijuma (approved):
LGTM

Zacharis278 (approved):
awesome looks good 👍

PR #144, opened 2024-04-12
fix: remove duplicate language in onboarding instructions

alangsto:
## [COSMO-233](https://2u-internal.atlassian.net/browse/COSMO-233)

A section of text in the submitted onboarding instructions was duplicated. I believe this may have been an oversight when copying over text for the instructions.

### DIFF
--- a/src/instructions/onboarding_exam/SubmittedOnboardingExamInstructions.jsx
+++ b/src/instructions/onboarding_exam/SubmittedOnboardingExamInstructions.jsx
@@ -27,7 +27,7 @@ const SubmittedOnboardingExamInstructions = () => {
       </h3>
       <p>
         <FormattedMessage
-          id="exam.SubmittedProctoredExamInstructions.text1"
+          id="exam.SubmittedProctoredExamInstructions.text4"
           defaultMessage={'If you do not have an onboarding profile with the system, Verificient '
           + 'will review your submission and create an onboarding profile to grant you access to '
           + 'proctored exams. Onboarding profile review can take 2+ business days.'}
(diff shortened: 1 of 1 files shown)


src/instructions/onboarding_exam/SubmittedOnboardingExamInstructions.jsx
```diff
@@ -51,14 +51,6 @@ const SubmittedOnboardingExamInstructions = () => {
       <p>
         <FormattedMessage
           id="exam.SubmittedProctoredExamInstructions.text4"
-          defaultMessage={'If you do not have an onboarding profile with the system, Verificient '
```
MichaelRoytman, line 54: This makes sense to me, but I have a concern about the ID.

Because we're removing the string with the ID `exam.SubmittedProctoredExamInstructions.text5`, I looked around for where it's used and where the string with ID `exam.SubmittedProctoredExamInstructions.text1` is used, because it'll be the ID of the string on the page. I noticed that `exam.SubmittedProctoredExamInstructions.text1` is used to identify two unique strings in this library. I'm not sure what happens when two IDs are used for two different strings - how does the Open edX translations code decide which string to send to Transifex for that ID? At least in [this locale, the other string is associated with this ID](https://github.com/openedx/openedx-translations/blob/main/translations/frontend-lib-special-exams/src/i18n/messages/es_ES.json#L40). I think it would be safer to keep this string with the ID `exam.SubmittedProctoredExamInstructions.text5` because it's unique. Do you see any issues with keeping this ID instead?

MichaelRoytman (changes requested):

src/instructions/onboarding_exam/SubmittedOnboardingExamInstructions.jsx
alangsto: No, that makes a lot of sense! I will change this back to using `exam.SubmittedProctoredExamInstructions.text5`

alangsto: I see what you're saying! Sorry I didn't catch that before. I will update the ID for `text1` to be `text4`, and then update `text4` to `text5`

MichaelRoytman (approved):
Thanks for bearing with me on that!

PR #147, opened 2024-07-11
fix: correct exam entry check not called

Zacharis278:
1. Resolved issue where the exam entry check for the Proctoring software was not being executed. Seems like we missed a dispatch() call here when doing some redux refactoring a few months ago and this broke.
2. Fixes an edge case where if the backend moves to a terminated state in a way that isn't caused by a client error the frontend does not show the correct interstitial message.

2U-internal Issue: [COSMO-291](https://2u-internal.atlassian.net/browse/COSMO-291)

### DIFF
--- a/src/data/thunks.js
+++ b/src/data/thunks.js
@@ -284,8 +284,9 @@ export function pollAttempt(url) {
       dispatch(setActiveAttempt({
         activeAttempt: updatedAttempt,
       }));
-      if (data.status === ExamStatus.SUBMITTED) {
+      if ([ExamStatus.SUBMITTED, ExamStatus.ERROR].includes(data.status)) {
         dispatch(expireExamAttempt());
+        updateAttemptAfter(exam.course_id, exam.content_id)(dispatch);
       }
     } catch (error) {
       handleAPIError(error, dispatch);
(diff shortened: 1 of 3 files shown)


alangsto (approved):
🚀

src/data/thunks.js
```diff
@@ -284,8 +284,9 @@ export function pollAttempt(url) {
       dispatch(setActiveAttempt({
         activeAttempt: updatedAttempt,
       }));
-      if (data.status === ExamStatus.SUBMITTED) {
+      if ([ExamStatus.SUBMITTED, ExamStatus.ERROR].includes(data.status)) {
         dispatch(expireExamAttempt());
+        updateAttemptAfter(exam.course_id, exam.content_id)(dispatch);
```
rijuma, line 289: I'm not sure I understand this call, shouldn't this be dispatched as:
```javascript
dispatch(updateAttemptAfter(exam.course_id, exam.content_id))
```
like the rest of the actions?

Zacharis278: this thing is a bit weird, it gets passed dispatch since this thunk calls to run another thunk passed to it as a third argument and than 'update the exam after'. We just need to do the after part so no third argument here. You can see the other uses of this throughout the file as an example tho.

rijuma: My bad. I was not used to this way of using the thunk.

PR #157, opened 2024-11-07
feat: hotfix to post-submit page for new ida

ilee2u:
https://2u-internal.atlassian.net/browse/COSMO-530

After completing a Practice Proctoring exam with an LTI provider. Reloading the page or navigating to the Dashboard and back to the exam an emtpy-ish page is shown instead of the usual landing.

I found that this is due to us not having a component that loads for Practice Exams for attempts in the "verified" or "rejected" state. I copied over the Verified/RejectedProctoredExamInstructions components to be loaded in the case that the exam's type is `PRACTICE` & we're using the new exams ida as a hotfix.

### DIFF
--- a/src/instructions/Instructions.test.jsx
+++ b/src/instructions/Instructions.test.jsx
@@ -364,6 +364,29 @@ describe('SequenceExamWrapper', () => {
     }
   });
 
+  it('Shows correct rejected practice exam instructions when attempt is rejected', () => {
+    store.getState = () => ({
+      specialExams: Factory.build('specialExams', {
+        exam: Factory.build('exam', {
+          is_proctored: true,
+          type: ExamType.PRACTICE,
+          attempt: Factory.build('attempt', {
+            attempt_status: ExamStatus.REJECTED,
+          }),
+        }),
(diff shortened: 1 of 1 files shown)


src/instructions/VerifiedInstructions.jsx
```diff
@@ -9,10 +9,12 @@ import Footer from './proctored_exam/Footer';
 const VerifiedExamInstructions = ({ examType }) => {
   const renderInstructions = () => {
     switch (examType) {
-      case ExamType.ONBOARDING:
-        return <VerifiedOnboardingExamInstructions />;
       case ExamType.PROCTORED:
         return <VerifiedProctoredExamInstructions />;
+      case ExamType.PRACTICE:
+        return <VerifiedProctoredExamInstructions />;
```
rijuma, line 15: Nit: Just for clarity, this could be written as:
```suggestion
      case ExamType.PROCTORED:
      case ExamType.PRACTICE:
        return <VerifiedProctoredExamInstructions />;
```

schenedx (approved):
👍

PR #160, opened 2025-02-20
chore: add logError for js worker call failures

ilee2u:
TICKET: https://2u-internal.atlassian.net/browse/COSMO-637

Adds a "logError" that we will use to track failures when making request to the JS Worker for proctoring (i.e. with the Proctortrack software, as that's the only proctoring provider we use a JS Worker with).

### DIFF
--- a/src/data/messages/handlers.js
+++ b/src/data/messages/handlers.js
@@ -1,3 +1,4 @@
+import { logError } from '@edx/frontend-platform/logging';
 import actionToMessageTypesMap from './constants';
 
 function createWorker(url) {
@@ -26,6 +27,7 @@ export function workerPromiseForEventNames(eventNames, workerUrl) {
           resolve();
         } else {
           reject(e.data.error);
+          logError('workerPromiseForEventNames failed with the following error:', e.data.error);
         }
       };
       proctoringBackendWorker.addEventListener('message', responseHandler);


src/data/messages/handlers.js
```diff
@@ -26,6 +27,7 @@ export function workerPromiseForEventNames(eventNames, workerUrl) {
           resolve();
         } else {
           reject(e.data.error);
+          logError('workerPromiseForEventNames failed with the following error:', e.data.error);
```
varshamenon4, line 30: Nice! I like how you include the error. Is there any other information that could be useful here when debugging?

varshamenon4 (approved):
lgtm! Just one q about the error log itself but not a blocker!

src/data/messages/handlers.js
ilee2u: Not sure what could be useful outside the error data. Though if this error goes off, then it's likely that the errors in the backend ([here](https://github.com/openedx/edx-proctoring/blob/73c7f55e2be91324fa07fec6e6ac0a667fdd8412/edx_proctoring/backends/rest.py#L116-L128) and [here](https://github.com/openedx/edx-proctoring/blob/73c7f55e2be91324fa07fec6e6ac0a667fdd8412/edx_proctoring/apps.py#L61-L68)) are likely going off as well, so if we need more details we can always check there.

PR #164, opened 2025-06-02
fix: ensure full-height layout

wgu-jesse-stewart:
Fixes https://github.com/openedx/frontend-app-learning/issues/1695
Related PR: https://github.com/openedx/frontend-app-learning/pull/1724

This PR simply moves `<div className="d-flex flex-column justify-content-center"> ` from the Exam component. to the learning MFE.

Summary:
When content within a sequence is shorter than the height of the browser viewport, the .sequence-container > .outline-sidebar-wrapper element does not expand appropriately. This causes the course outline sidebar to be partially or completely hidden from view.

Steps to Reproduce:

Open a course sequence that contains very little content (e.g., a short paragraph or notice).

View the course in a tall browser window.

Observe that the sidebar navigation is cut off or hidden.

Expected Behavior:
The sidebar should remain fully visible and accessible, regardless of the amount of content in the main pane.

Actual Behavior:
When the content is short, the parent container height collapses to fit that content, which in turn restricts the sidebar’s height, hiding part of the navigation.

Screenshots:
image

image

Demo of fix:
https://github.com/user-attachments/assets/69ce8dae-e73b-487c-bc6a-862701b1ce8d

I think we should have a discussion about removing these wrapper divs entirely so that the layout can handle all the UI

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,7 @@
       "main"
     ]
   },
-  "exports": {
-    "import": "./dist/index.js"
-  },
+
   "repository": {
     "type": "git",
     "url": "git+https://github.com/edx/frontend-lib-special-exams.git"
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
       "main"
     ]
   },
-
+  "exports": {
+    "import": "./dist/index.js"
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/edx/frontend-lib-special-exams.git"
--- a/src/exam/Exam.jsx
+++ b/src/exam/Exam.jsx
@@ -95,7 +95,7 @@ const Exam = ({
   const sequenceContent = <>{children}</>;
 
   return (
-    <div className="d-flex flex-column flex-grow-1 justify-content-center">
+    <>
       {shouldShowMasqueradeAlert() && (
         <Alert variant="info" icon={Info} data-testid="masquerade-alert">
           <FormattedMessage
@@ -113,7 +113,7 @@ const Exam = ({
       {isTimeLimited && !originalUserIsStaff && !isGated
         ? <Instructions>{sequenceContent}</Instructions>
         : sequenceContent}
-    </div>
+    </>
   );
 };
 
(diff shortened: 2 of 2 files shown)


diana-villalvazo-wgu (approved):

holaontiveros (approved):

brian-smith-tcril (approved):
:shipit:

arbrandes:
Looks like the release workflow isn't working, for some reason.  I'll look into it.

brian-smith-tcril:
@arbrandes I think the config `semantic-release` uses doesn't mark commits as breaking based on just `!` in the type, and instead requires `BREAKING CHANGE` to be in the footer. It's not clear to me why it's saying
```
[3:54:41 PM] [semantic-release] [@semantic-release/commit-analyzer] › ℹ  The commit should not trigger a release
```
though.

arbrandes:
@brian-smith-tcril, yeah, I tried with `BREAKING CHANGE` instead, and it just worked.  👍🏼

wgu-jesse-stewart:
> @brian-smith-tcril, yeah, I tried with `BREAKING CHANGE` instead, and it just worked. 👍🏼

I think it would be useful to add this to the README. But I want to be sure it is accurate first.

## Automatic Releases

This repository uses [semantic-release](https://semantic-release.gitbook.io/) to automate versioning and publishing.

### Triggering a Release

Releases are triggered based on commit messages that follow [Conventional Commits](https://www.conventionalcommits.org/).

To automatically create a new release:

- Use one of the following commit types: `feat`, `fix`, or `chore`.
- Include a scope if relevant, for example:  
  `feat(layout): adjust height behavior`
- To trigger a **breaking change** and a **major version bump**, include a `BREAKING CHANGE:` footer in the commit message, like so:

  ```text
  feat: adjust full-height layout

BREAKING CHANGE: layout structure was modified and may break consumers relying on previous DOM structure
```

**Note:** Just using feat!: does not trigger a breaking change unless BREAKING CHANGE: is present in the footer. This is due to the current semantic-release configuration.

